PR #166, opened 2017-01-23
Update Validator.php

vvval:
If use `MyChecker::class . '::validationMethod'` without adding `MyChecker::class` alias into validation config then we have 2 problems:

1. We can't use default messages
2. We can't use `$this->validator()` inside of a checker


wolfy-j:
Won't merge without tests.

PR #167, opened 2017-02-06
pull request validator check without alias

vvval:
Ability to add validation rule without registering checker alias like 'myCheckers\MyChecher::myRule' with nex advantages:

1. Default messages will be available on check fail.
2. `$this->getValidator()` will be available inside of a checker.


wolfy-j:
```php
if ($this->config->hasChecker($condition[0])) {
 
                     $checker = $this->getChecker($condition[0]);
                     $result = $checker->check($condition[1], $value, $arguments);
 
                     if ($result === false) {
                         //To let validation() method know that message should be handled via Checker
                         return $checker;
                     }
  
                      return $result;
                  }
```

I think it's better to alter this section hasChecker and getChecker instead of making new piece of code which is doing the same thing.

Issue #180, opened 2017-04-28
[Validation package] Don't validate the same data set twice (details inside)


adelowo:
`validate` is called in `getErrors`  [here](https://github.com/spiral/spiral/blob/develop/source/Spiral/Validation/Validator.php#L206) and `isValid` [here](https://github.com/spiral/spiral/blob/develop/source/Spiral/Validation/Validator.php#L168). 

The goal is to remove code duplication..  Both `isValid` and `getErrors` calls the validate method before returning a response to the caller.. This leads to performing the same chore twice. 

For instance, say i call `isValid` (validation runs here), if i get a falsy value, i obviously would want to call `getErrors` to show the user who went wrong. On calling getErrors,the data is re-validated before the errors are returned.

adelowo:
Kind of related --- > 

I also think the `validate` method should be exposed on the `ValidatorInterface`[here](https://github.com/spiral/spiral/blob/develop/source/Spiral/Validation/ValidatorInterface.php) as it should lead to better usability (since it actual does the validation not `isValid`.... I tried working on this (issue) but the tests are broken, while i have fixed most, i have hit a stumbling block with the `ValidatesEntity`

wolfy-j:
Fair point, we had error cache in 0.8 version, but due refactoring this micro-optimization is gone. I'll take a look if test suite ready to handle this change.

adelowo:
I don't think the test suite is ready for that.. I did it and everything got broken..

Like i said earlier, i am still yet to fix the tests for `ValidatesEntity`

adelowo:
I would have to wait for you to check this before i continue working on this issue

wolfy-j:
Yes, i will take care of it next week after holidays. :)

wolfy-j:
The trick here is to properly invalidate cache for validation errors.

adelowo:
Oh, I would take a look at this tomorrow (or this night if am chanced) again

adelowo:
The tests are still broken even after merging `feature/validate-entity` and exposing the 1Validate` method.. `ValidatesEntityTest` is the real culprit here...

Would look into this properly tomorrow.

wolfy-j:
I can handle it i think, but only after weekends :)

adelowo:
Alright then

wolfy-j:
see #183

PR #181, opened 2017-04-28
[Validation Package] Allow validation data to be set directly from an EntityInterface

adelowo:
This PR allows validation data to be set from a `DataEntity` (or `EntityInterface` to be precise)....

Usecase = >

- Currently 

```php

$data = [];
$data[self::ARGS_FULL_NAME] = $this->argument(self::ARGS_FULL_NAME);
$data[self::ARGS_MONIKER] = $this->argument(self::ARGS_MONIKER);
$data[self::ARGS_EMAIL] = $this->argument(self::ARGS_EMAIL);

$transformedData = new SomeDataEntity($data);

$val->setData($transformedData->getFields());

//Do something else with $transformedData
```

- This PR =>

```php
$validator = $validator->withDataEntity($transformedData);
```

Just sugar syntax, but i have found myself this past hour filling an `EntityInterface` and passing it to the validator.

### DIFF
--- a/source/Spiral/Validation/Validator.php
+++ b/source/Spiral/Validation/Validator.php
@@ -127,6 +127,14 @@ class Validator extends Component implements ValidatorInterface, LoggerAwareInte
     /**
      * {@inheritdoc}
      */
+    public function withDataEntity(EntityInterface $entity) : ValidatorInterface
+    {
+        return $this->setData($entity->getFields());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function setRules(array $rules): ValidatorInterface
     {
         if ($this->rules == $rules) {
@@ -243,7 +251,6 @@ class Validator extends Component implements ValidatorInterface, LoggerAwareInte
     {
         $this->errors = [];
         foreach ($this->rules as $field => $rules) {
-
             foreach ($rules as $rule) {
                 if (isset($this->errors[$field])) {
                     //We are validating field till first error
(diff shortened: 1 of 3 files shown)


adelowo:
I don't get why coverage reduced :1234:

wolfy-j:
Few comments:
1) "with" prefix usually used to represent immutable methods, will lead to confusion
2) I'm not sure if it's a good idea to couple validator (and validator interface) an entity as this is parts of different domain

ValidatesEntity combines them together as implementation, but joining this two abstractions on highest level will create higher coupling. 

Another option is too make setData method accept entity directly. Let me think about it till next week.

adelowo:
Alright, but at worse, `setData` should be able to work with an array or an `EntityInterface`

wolfy-j:
Yes it should. http://prntscr.com/f261nc If it's not, you just found a bug :)

wolfy-j:
Actually, it should be working with ArrayAccess which is implemented by all entities but not in the interface.  I will correct that.

adelowo:
I missed noticing it accepts `\ArrayAccess`..

wolfy-j:
Well, EntityInteface is not necessary ArrayAccess so you had a fair point. I created a new branch, let's wait for CI.

wolfy-j:
https://github.com/spiral/spiral/tree/feature/validate-entity

wolfy-j:
BTW, there is a slight chance that cache will work as well now, Validator itself caches errors IF data not changed. Regarding #180

adelowo:
Ok, i am just going to pull that branch and test it out

wolfy-j:
Cool, i included your test as well :)

adelowo:
Closed by #182

PR #211, opened 2019-07-29
Add environment configuration for worker relay

myavchik:


wolfy-j:
Thank you!

wolfy-j:
I've added the fallback to when no ENV is defined. I think ideally I would like to configure it automatically via RR.

myavchik:
Sure, auto configuration via RR is better than duplicating code in .env and .rr.yaml.

wolfy-j:
Can you please create a ticket in RR, we can add it to the next version.

myavchik:
I just checked RR code and found what it already setting env RR_RELAY. Sorry, my lapse. I will create new PR with RR_WORKER renamed to RR_RELAY.

wolfy-j:
Oh right, I've forget about it. I'm preparing the release now, I'll change it by myself.

PR #232, opened 2020-01-14
Json payload allow custom content type

R3VoLuT1OneR:
Add new configuration in http config, that will allow custom MIME types for JSON payloads.

For example if someone wants to use `text/json` as a MIME types for JSON payloads, he can simply extend HTTP config to allow such "content-type" to be parsed as JSON.

Inspired by willing to implement JSON:API with Spiral framework, but have some doubts about need of JsonPayloadMiddleware be a part of the framework code, as there is different libraries solving it already. Example:
https://github.com/middlewares/payload#jsonpayload


R3VoLuT1OneR:
Didn't put attention that modified `spiral/http` to make it work, will create another PR later.

PR #249, opened 2020-03-26
Fix GRPC Generate Path Argument

matthewhall-ca:
Currently the GRPC Generate command doesn't actually use it's Path argument


wolfy-j:
Thank you!

Issue #252, opened 2020-04-12
Possible cast error ?


ncou:
Hi,

Still looking at the code, i think there could be a cast error in this piece of code : 

https://github.com/spiral/framework/blob/master/src/Command/Router/ListCommand.php#L139

The "action" parameter in the class "Action" could be a string but also an array (as indicated in the docbloc : https://github.com/spiral/router/blob/master/src/Target/Action.php#L37).

You will surely have an cast array to string error. Perhaps adding a "join("|", $action)" could prevent this potential error.

Keep up the good work.

wolfy-j:
Thank you for the feedback. PR will be appreciated or it will have to wait for when we will have the capacity for this task.

ncou:
I can push a PR.
I was thinking to add this piece of code to handle the case when the var is a string or an array :+1: 

```sprintf("%s->%s", 'controler', join('|', (array) $action))```php

Is this ok with you ?
I also see there is no test for the command class, so i will only modify the ListCommand file.

wolfy-j:
This looks OK. This is debugging command so testing it wasn't in priority.

ncou:
The PR is done, looking in the GithubAction, the build is green. Let me know if there is another thing to do.

ncou:
PR Merged. Have a good day.

PR #253, opened 2020-04-13
fix display of 'action' value for the Target Action when the value is an array

ncou:
fix issue : https://github.com/spiral/framework/issues/252


wolfy-j:
Thank you!

PR #264, opened 2020-05-20
release 2_4_18

48d90782:
update rr to 1.8.1
update broadcast to version w\o hash


bin/spiral
```diff
@@ -9,8 +9,8 @@ declare(strict_types=1);
 
 foreach ([
              __DIR__ . '/../../../autoload.php',
-             __DIR__ . '/../vendor/autoload.php',
-             __DIR__ . '/vendor/autoload.php'
+             __DIR__ . '/../vendor_php/autoload.php',
```
wolfy-j, line 12: This is not correct. The vendor path will only change for local development, the actual project will still use vendor.

src/Bootloader/Security/ValidationBootloader.php
```diff
@@ -114,7 +114,7 @@ public function boot(TokenizerBootloader $tokenizer): void
             ]
         );
 
-        $tokenizer->addDirectory(directory('vendor') . 'spiral/validation/src/');
```
wolfy-j, line 117: Same

src/Framework/Kernel.php
```diff
@@ -57,7 +57,7 @@ protected function mapDirectories(array $directories): array
                 'public'    => $directories['root'] . '/public/',
 
                 // vendor libraries
-                'vendor'    => $directories['root'] . '/vendor/',
```
wolfy-j, line 60: Same

tests/bootstrap.php
```diff
@@ -12,4 +12,4 @@
 ini_set('display_errors', '1');
 
 //Composer
-require dirname(__DIR__) . '/vendor/autoload.php';
```
wolfy-j, line 15: This one can stay with vendor_php

bin/spiral
48d90782: done

src/Framework/Kernel.php
48d90782: done

tests/bootstrap.php
48d90782: done

src/Bootloader/Security/ValidationBootloader.php
48d90782: done

PR #267, opened 2020-06-05
Allow json requests with empty body

rauanmayemir:
This will loosen the restriction on sending requests with empty body.
The thinking behind this is that `Content-Type` header usually set codebase-wide and being rigorous in this regard doesn't seem to give any benefits.


wolfy-j:
Thank you!

PR #275, opened 2020-07-31
Register present and absent validation conditions

rauanmayemir:


vvval:
Shouldn't you bump validation dependency in composer.json then?

rauanmayemir:
@vvval Oops. What was the reason for locking in the patch version? Should we make it `^1.4`?

wolfy-j:
Yes, it should be ^1.4.4 (where you introduced new class).

PR #277, opened 2020-08-14
Monrepo initialization

SerafimArts:
- The sources have been temporarily moved into `src/Framework` (namespace not changed; no polyfill required).
- Tests namespace was brought to canonical form `Spiral\Tests\***`.
  - All namespace typos (inside tests) fixed.
- All composer files are converted to canonical form.
- Version `2.6.x-dev` has been created.

Please note that this is only one `spiral/core` package. After merge this PR it will be tagged as readonly.


wolfy-j:
Can you check why tests are failing?

SerafimArts:
> Can you check why tests are failing?

Its using assertions from phpunit 9.0.

Issue #297, opened 2020-09-18
Empty cycle schema may break app launch


vvval:
In some specific cases, when a repository is injected via DI and there's an empty-array schema, the app may fail with the next error:
```
bash-5.0$ php app.php cycle
 [Cycle\ORM\Exception\ORMException]                                        
 Unable to find Entity role for repository App\Repository\OriginRepository 
in /var/www/vendor/spiral/framework/src/Cycle/RepositoryInjector.php:56

Exception Trace:
 Spiral\Cycle\RepositoryInjector->createInjection() at /var/www/vendor/spiral/framework/src/Cycle/RepositoryInjector.php:56
 Spiral\Cycle\RepositoryInjector->createInjection() at /var/www/vendor/spiral/core/src/Container.php:551
 Spiral\Core\Container->createInstance() at /var/www/vendor/spiral/core/src/Container.php:422
 Spiral\Core\Container->autowire() at /var/www/vendor/spiral/core/src/Container.php:147
 Spiral\Core\Container->make() at /var/www/vendor/spiral/core/src/Container.php:133
 Spiral\Core\Container->get() at /var/www/vendor/spiral/core/src/Container.php:245
 Spiral\Core\Container->resolveArguments() at /var/www/vendor/spiral/core/src/Container.php:575
 Spiral\Core\Container->createInstance() at /var/www/vendor/spiral/core/src/Container.php:422
 Spiral\Core\Container->autowire() at /var/www/vendor/spiral/core/src/Container.php:147
 Spiral\Core\Container->make() at /var/www/vendor/spiral/core/src/Container.php:133
 Spiral\Core\Container->get() at /var/www/vendor/spiral/annotated-routes/src/RouteGroup.php:99
 Spiral\Router\RouteGroup->addMiddleware() at /var/www/app/src/Bootloader/ErrorCausingBootloader.php:33
 App\Bootloader\Api\RoutingBootloader->boot() at n/a:n/a
 ReflectionMethod->invokeArgs() at /var/www/vendor/spiral/boot/src/BootloadManager.php:124
 Spiral\Boot\BootloadManager->initBootloader() at /var/www/vendor/spiral/boot/src/BootloadManager.php:98
 Spiral\Boot\BootloadManager->boot() at /var/www/vendor/spiral/boot/src/BootloadManager.php:66
 Spiral\Boot\BootloadManager->Spiral\Boot\{closure}() at /var/www/vendor/spiral/core/src/Container.php:282
[...]

PR #298, opened 2020-09-18
#297 add possibility to regenerate schema if the memory file is empty

vvval:


SerafimArts (approved):

src/Framework/Bootloader/Cycle/SchemaBootloader.php
```diff
@@ -107,15 +110,23 @@ public function getGenerators(): array
     protected function schema(MemoryInterface $memory): SchemaInterface
     {
         $schema = $memory->loadData('cycle');
-        if (is_null($schema)) {
+        if (empty($schema)) {
```
SerafimArts, line 113: I think that it can be rewritten as:
```php
$schema = $memory->loadData('cycle');

if ($schema) {
    return new Schema($schema);
}

$schema = (new Compiler())->...
```

Early return will help you get rid of the extra nesting level.

P.S. And the schemа generation can be taken out in a separate private method, then you get result like:
```php
$schema = $memory->loadData('cycle');

return $schema ? new Schema($schema) : $this->createDefaultSchema();
```

SerafimArts (changes requested):

src/Framework/Bootloader/Cycle/SchemaBootloader.php
```diff
@@ -107,15 +110,23 @@ public function getGenerators(): array
...
             );
 
+            if (empty($schema)) {
+                $schema = self::EMPTY_SCHEMA;
+            }
+
             $memory->saveData('cycle', $schema);
         }
 
+        if ($schema === self::EMPTY_SCHEMA) {
```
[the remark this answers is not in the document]
vvval: Schema should be an array of arrays. I've changed so `:empty:` now is a whole value, not an array

src/Framework/Bootloader/Cycle/SchemaBootloader.php
vvval: Refactored the whole operation

SerafimArts (approved):

Issue #294, opened 2020-09-02
Container isn't compatible with PHP 8


samdark:
https://github.com/spiral/framework/blob/master/src/Core/src/Container.php#L197 is not compatible with PHP 8 giving

```
Method ReflectionParameter::getClass() is deprecated
```

It could be replaced with something like:

```php
$name = $param->getType() && !$param->getType()->isBuiltin() ? new ReflectionClass($param->getType()->getName()) : null;
```

samdark:
https://github.com/yiisoft/yii-cycle/runs/1058384659#step:12:83

wolfy-j:
I think @SerafimArts already addresses this in upcoming release.

SerafimArts:
@samdark I already researched this issue a while ago. The example you specified is not entirely correct, since does not work with union types.

samdark:
Yes, that I know. Union types should be handled like https://github.com/yiisoft/injector/blob/master/src/Injector.php#L198

roxblnfk:
@SerafimArts when are you planning to resolve this issue?

SerafimArts:
@roxblnfk At the moment I have slightly different plans, but this issue is a priority one.

As far as I understand, this problem is currently a blocker for you?

roxblnfk:
> As far as I understand, this problem is currently a blocker for you?

No, just the tests are red

SerafimArts:
> No, just the tests are red

Only when running on PHP 8 =) This is the reason why this version is not in CI yet: https://github.com/spiral/framework/blob/master/.github/workflows/main.yml#L63

DanRaduDragomir:
Seems like this reappeared in the latest release:

```log
ERRO[0000] [grpc]: [grpc]: unable to connect to worker:  [ErrorException]                                     
Method ReflectionParameter::getClass() is deprecated 
in /code/vendor/spiral/php-grpc/src/Method.php:68
```

and yes, looking in that file we will find:
```php
public static function match(\ReflectionMethod $method): bool
    {
        if ($method->getNumberOfParameters() != 2) {
            return false;
        }

        $ctx = $method->getParameters()[0]->getClass();
        $in = $method->getParameters()[1]->getClass();
[...]
```

PR #300, opened 2020-09-19
Fixed PHP 8 deprecation error when getting ReflectionClass from argument type hint

SerafimArts:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌
| Issues        | #294 

During attempt to resolve the parameter, the `$prarmeter->getClass()` method was called, which was removed from PHP8 due to the addition of union types: https://github.com/JetBrains/phpstorm-stubs/blob/master/Reflection/ReflectionParameter.php#L107-L116

1) The logic was rewritten and now when reading the argument's typhint, it is read correctly.
2) In the case of a resolving dependency from a union type hint, an error about the ambiguity of autowiring will occur. The functionality of correct handling of such situations requires a separate discussion.


SerafimArts:
@wolfy-j @vvval Because a lot of time has passed and no one has revised, then I merge.

This PR is required to repair tests: https://github.com/spiral/framework/runs/1393569833?check_suite_focus=true

PR #384, opened 2021-01-19
Fix execution of EnvironmentCollector::populate outside of DispatcherInterface context

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌

Executing `EnvironmentCollector::populate` outside of `DispatchInterface` context will result in exception on `if ($this->container->has(DispatcherInterface::class)) {` line.

Currently this exception is caught in `ErrorHandlerMiddleware::getOptional`.

AFAIK `container::get()` can not return non-object result which renders this `if` statement useless.


wolfy-j (approved):

wolfy-j:
Looks good to me!

PR #385, opened 2021-01-20
Fix GridFactoryInterface -> GridFactory binding

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌

Current `GridFactory::class => GridFactory::class` declaration doesn't really do anything except declaring GridFactory a singleton. I guess it was intended to be written as this fix suggests

### DIFF
--- a/src/Bridge/DataGrid/src/Bootloader/GridBootloader.php
+++ b/src/Bridge/DataGrid/src/Bootloader/GridBootloader.php
@@ -14,6 +14,7 @@ namespace Spiral\DataGrid\Bootloader;
 use Psr\Container\ContainerInterface;
 use Spiral\Boot\Bootloader\Bootloader;
 use Spiral\Config\ConfiguratorInterface;
+use Spiral\DataGrid\GridFactoryInterface;
 use Spiral\Database\DatabaseInterface;
 use Spiral\DataGrid\Compiler;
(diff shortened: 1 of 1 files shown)


SerafimArts:
I have a suspicion that BC is here.

Registering an interface instead of implementing it is correct. However, in the case that someone is already using the implementation in the form of a service location `->get(GridFactory::class)`, then after the patch everything will break =\

It will be necessary to come up with a mechanism with which it will be possible to mark services as deprecated, like: `$container->depreacte(GridFactory::class, 'Since 2.x. Please use GridFactoryInterface instead')`

SerafimArts (approved):

vladgorenkin:
Not really sure why `$container->get(FilterFactory::class)` would stop working, but updated PR according to your comment

SerafimArts:
@wolfy-j, can you please confirm or deny my thesis?

wolfy-j (approved):

wolfy-j:
Yes,

some people might have been using class dependency. This PR is good now!

wolfy-j:
Thank you, @vladgorenkin

SerafimArts:
@vladgorenkin Sorry to wait. There has been too much work lately =)

Issue #377, opened 2020-12-23
ExceptionHandler does not account for error_reporting setting


vladgorenkin:
### Description

Current error handler does not take into account `error_reporing` setting.  For instance if I set `error_reporting((E_ALL | E_STRICT) ^ E_USER_DEPRECATED)` in app.php to avoid fatals on deprecations application should not throw exceptions on `\trigger_error('...', \E_USER_DEPRECATED)`, but it currently does.

Solution would be adding a guard to `Spiral\Boot\ExceptionHandler::handleError`:

```php
public static function handleError($code, $message, $filename = '', $line = 0): void
{
    if (!(error_reporting() & $code)) {
        return;
    }

    throw new \ErrorException($message, $code, 0, $filename, $line);
}
```

### How To Reproduce

```php
// app.php

...
error_reporting((E_ALL | E_STRICT) ^ E_USER_DEPRECATED);
...
```

```php
// any code after set_error_handler is called in Spiral\Boot\ExceptionHandler

trigger_error('Test error', E_USER_DEPRECATED);
```

### Additional Info

| Q                   | A
| ------------------- | ---
| Framework Version   | 2.6.3
| PHP version         | 7.4.11
| Operating system    | Linux

PR #386, opened 2021-02-01
Respect `error_reporting` setting in ExceptionHandler::handleError

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ I guess
| New feature?  | ❌
| Issues        | #377

For description see related issue #377. This change should not break anything, no new exceptions will be thrown after merge.

### DIFF
--- a/src/Boot/src/ExceptionHandler.php
+++ b/src/Boot/src/ExceptionHandler.php
@@ -75,6 +75,10 @@ final class ExceptionHandler
      */
     public static function handleError($code, $message, $filename = '', $line = 0): void
     {
+        if (!(error_reporting() & $code)) {
+            return;
+        }
+
         throw new \ErrorException($message, $code, 0, $filename, $line);
     }
 


src/Boot/src/ExceptionHandler.php
```diff
@@ -75,6 +75,10 @@ public static function handleShutdown(): void
      */
     public static function handleError($code, $message, $filename = '', $line = 0): void
     {
+        if (!(error_reporting() & $code)) {
```
SerafimArts, line 78: This is side effect. 

I think implementation like this should be prefer:
```php
public function __construct(int $reportingLevel = null)
{
    $this->reportingLevel = $reportingLevel ?? \error_reporting();
}

/// ...
if (!($this->reportingLevel & $code)) {
    return;
}
```
What do you think?

SerafimArts (approved):
I will approve, because there is nothing critical =)

src/Boot/src/ExceptionHandler.php
vladgorenkin: I think side effect is expected here, though I agree generally it's not a good practice. `error_reporting()` operates as a global variable setter/getter, so if we cache it may confuse some future users. For example, if someone wants to turn off exceptions for a certain piece of code he should be able to do it by changing error_reporting twice.

I took a quick look at how other packages implement errorHandlers. Here is monolog:
```
    public function handleError($code, $message, $file = '', $line = 0, $context = [])
    {
        if ($this->handleOnlyReportedErrors && !(error_reporting() & $code)) {
            return;
        }
   ...
```

Here is sentry:
```
    private function handleError(int $level, string $message, string $file, int $line, ?array $errcontext = []): bool
    {
        if (0 === (error_reporting() & $level)) {
            $errorAsException = new SilencedErrorException(self::ERROR_LEVELS_DESCRIPTION[$level] . ': ' . $message, 0, $level, $file, $line);
        } else {
            $errorAsException = new \ErrorException(self::ERROR_LEVELS_DESCRIPTION[$level] . ': ' . $message, 0, $level, $file, $line);
        }
    ...
```

wolfy-j:
Cool, let's merge it when you'll be free Kirill.

PR #394, opened 2021-03-09
Prevent dumprr from outputing html-wrapped results

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌

Currently `dumprr` uses HtmlRenderer to render results which does not make much sense when outputing to stderr. This patch fixes it.

Related discussion: https://github.com/spiral/framework/commit/0e5a4bc38eeaa65a1f6250490ea6c1853564096d#r47916628


SerafimArts:
Thanks a lot!

PR #401, opened 2021-04-08
Use NamedArgumentConstructorAnnotation as an extra condition in Instantiator\Factory

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️

Using NamedArgumentConstructorAnnotation interface as a marker is deprecated since doctrine/annotations:1.12, so here is an additional check

### DIFF
--- a/src/Attributes/src/Internal/Instantiator/Factory.php
+++ b/src/Attributes/src/Internal/Instantiator/Factory.php
@@ -11,7 +11,9 @@ declare(strict_types=1);
 
 namespace Spiral\Attributes\Internal\Instantiator;
 
+use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 
 class Factory implements InstantiatorInterface
 {
@@ -26,12 +28,18 @@ class Factory implements InstantiatorInterface
     private $named;
 
     /**
+     * @var AnnotationReader
+     */
+    private $annotationReader;
+
+    /**
      * Factory constructor.
      */
     public function __construct()
     {
         $this->doctrine = new DoctrineInstantiator();
         $this->named = new NamedArgumentsInstantiator();
+        $this->annotationReader = new AnnotationReader();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


SerafimArts (approved):

src/Attributes/src/Internal/Instantiator/Factory.php
```diff
@@ -25,13 +27,19 @@ class Factory implements InstantiatorInterface
...
+    private $annotationReader;
+
     /**
      * Factory constructor.
      */
     public function __construct()
     {
         $this->doctrine = new DoctrineInstantiator();
         $this->named = new NamedArgumentsInstantiator();
+        $this->annotationReader = new AnnotationReader();
```
SerafimArts, line 42: An annotation reader cannot be instantiated in internal classes because this is an optional dependency. The reader type must be defined by the user himself.

src/Attributes/src/Internal/Instantiator/Factory.php
```diff
@@ -55,6 +63,10 @@ public function instantiate(\ReflectionClass $attr, array $arguments, string $co
      */
     private function isNamedArguments(\ReflectionClass $class): bool
     {
-        return \is_subclass_of($class->getName(), NamedArgumentConstructorAnnotation::class);
+        return is_subclass_of($class->getName(), NamedArgumentConstructorAnnotation::class)
+            || (
+                class_exists(NamedArgumentConstructor::class)
+                && null !== $this->annotationReader->getClassAnnotation($class, NamedArgumentConstructor::class)
```
SerafimArts, line 69: Instantiation by annotation cannot be done on attributes.

Moreover, this annotation is not an attribute, so it cannot be read by an attribute reader.

That is, the case is not taken into account:
```php
#[NamedArgumentConstructor]
#[\Attribute(\Attribute::TARGET_ALL)]
class MyAttribute
{
    public function __construct(string $some) { ... }
}
```

SerafimArts:
The first approval came from my mistake. I have not looked or analyzed specific use cases for this functionality.

Note that the intantiator emulates the behavior of the doctrine constructor when using attributes for easier transition from the old syntax.

I think that in general, it makes sense to replace all the code with:

```php
// constructor
if (class_exists(DoctrineAnnotationReader::class)) {
    $this->reader = new SpiralAnnotationReader();
}

// isNamedArguments function
if (is_subclass_of($class->getName(), NamedArgumentConstructorAnnotation::class)) {
    return true;
}

if ($this->reader !== null) {
    // Try detect constructor type using annotation
}
```

?

SerafimArts:
Merged here: https://github.com/spiral/attributes/pull/2

P.S. This branch that you are using is not relevant yet, because the package is distributed separately. And although it is in release, it is not stable and has not been merged into the main framework branch.

vladgorenkin:
Thanks. Sorry for late response, haven't had a moment to fix PR this week

PR #419, opened 2021-06-08
Implement lazy-loading of cli commands

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ✔️
| New feature?  | ✔️

This PR implements lazy-loading of CLI commands to prevent unneeded commands from being constructed. This speeds up CLI initialization and solves situations when one command's dependency breaks all other commands (e.g. a command whose dependency uses result of ORM query will break migrate:init when no migrations were applied).

Lazy-loaded command declaration is done through upgrading `Command::NAME`'s visibility to public so locators can extract command's name (and optionaly description) from command without constructing it.

Example:

```php
class LazyCommand extends Spiral\Console\Command {
    public const NAME = 'lazy';

    public function __construct() {
        sleep(5); // this will only be run when command is executed, not on console initialization
    }
}
```


SerafimArts:
@vladgorenkin This breaks BC =(

In addition, I don't really want to be tied to agreements with "magic constants". However, I understand that this style has already been adopted in Spiral 2.x and should be followed.

vladgorenkin:
> It breaks BC

In what way? Dependency on LazyCommand? No problem in waiting till 2.9 though, no hurry here.

> In addition, I don't really want to be tied to agreements with "magic constants"

I absolutely agree, I do not like magic solutions either although it seemed to me as the most suitable taking into account existing logic. This solution was also partly inspired by Symfony's way of declaring lazy commands (by declaring $defaultName property static).

The problem is we need to extract name and description of a command to register it in Application so NAME and DESCRIPTION need to be public. If you have any better solution in mind please let me know.

SerafimArts:
> In what way? Dependency on LazyCommand?

Yep. This requires bringing up a version of the Symfony's components. There is no such thing now: https://github.com/spiral/framework/pull/419/checks?check_run_id=2772546255

SerafimArts:
@vladgorenkin Can you raise a version of the symfony dependencies to pass all the tests?

vladgorenkin:
✔️

SerafimArts:
This is not quite the correct way. When the build starts, everything will break.

The final composer file is built from all existing packages using `vendor/bin/monorepo-builder merge` command. You should fix the versions for all packages (inside `src` dir) and run this command.

Thanks anyway. This is not critical, I will fix everything when I do (merge) this PR.

SerafimArts (approved):

vladgorenkin:
Just for the sake of my own education tried to resolve conflicts and generate composer.json per your instructions. Would you have a look and tell me that everyhing is ok?

SerafimArts:
Yes, it's OK. 👍

We did not add this information to the documentation, since the case of version changes and merging of child packages is very rare. In the future, we may automate this process by CI so that the versions are all updated independently, but for now we have to do this manually.

SerafimArts:
Yes. 👍

I see that there are problems with `jobs` and `migrations`. This seems to be an automation problem, since these packages are optional and have not yet been versioned along with the framework. Thank you for noticing this and correcting it too.

Issue #402, opened 2021-04-16
Mail Message options are not serialized for queue


vladgorenkin:
## Description

Not sure if it is intended behavior but currently MessageInterface::$options are not serialized, so you can't use them when sending mails asynchronously

```
$message = (new Message(...))->setOptions(['foo' => 'bar']);
MailQueue->send($message);

// SomeRenderer.php
$message->getOptions(); // []
```

SerafimArts:
Merged (available since v2.8.3 release https://github.com/spiral/framework/releases/tag/v2.8.3)

PR #422, opened 2021-06-10
Serialize Mail::options for sending via queue

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌
| Issues        | #402


SerafimArts:
@vladgorenkin Thank you so much!

Could you also upload the changes to the 2.8 branch? In this case, this fix can be immediately committed under the new `2.8.3` relese tag. If the fix is not critical, then it is possible in the `master`, but in this case it will appear in the stable tag 2.9 (i.e. feature release) at the beginning of the next month.

SerafimArts (approved):

vladgorenkin:
Done. But it seems that 2.8 branch is behind v2.8.2.

PR #426, opened 2021-06-29
fix(distribution): always prefix file uri with slash

hustlahusky:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌
| Issues        |
| Docs PR       | 

always prefix file uri with slash, to prevent guzzle uri trigger an error, when resolving file from subdirectory

### DIFF
--- a/src/Distribution/src/Resolver/StaticResolver.php
+++ b/src/Distribution/src/Resolver/StaticResolver.php
@@ -72,11 +72,9 @@ class StaticResolver extends UriResolver
      */
     private function suffix(string $file): string
     {
-        $prefix = $this->host->getPath();
+        $prefix = \trim($this->host->getPath(), self::URI_PATH_DELIMITER);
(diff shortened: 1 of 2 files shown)


SerafimArts (approved):

hustlahusky:
> It seems like one could just add "`/`" before the `implode` function))

This will generate double slash if no prefix in host

> Can you please send this PR to branch 2.8, instead of the master?

I have changed base branch to 2.8. Is it okay, or I should cherrypick commit to new branch?

SerafimArts:
This is a bug fix, so it had to go to `v2.8.next`, and you sent it to the next release (to version `v2.9-dev`). 

Yep, you should send this commit separately. Now you propose to merge all the changes of the next version (2.9) into 2.8

SerafimArts:
👍🏻

hustlahusky:
Branch updated so it can be merged to 2.8

PR #435, opened 2021-07-27
Add array::count, array::range, array::shorter and array::longer rules

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ✔️ updated error messages for number::lower and number::higher
| New feature?  | ✔️

Allows to validate number of elements in array

### DIFF
--- a/src/Validation/src/Checker/ArrayChecker.php
+++ b/src/Validation/src/Checker/ArrayChecker.php
@@ -9,6 +9,16 @@ use Spiral\Validation\ValidationInterface;
 
 class ArrayChecker extends AbstractChecker
 {
+    /**
+     * {@inheritdoc}
+     */
+    public const MESSAGES = [
+        'count' => '[[Number of elements must be exactly {1}.]]',
+        'longer' => '[[Number of elements must be equal to or greater than {1}.]]',
+        'shorter' => '[[Number of elements must be equal to or less than {1}.]]',
+        'range' => '[[Number of elements must be between {1} and {2}.]]',
+    ];
+
     /** @var ValidationInterface */
     private $validation;
 
@@ -30,5 +40,43 @@ class ArrayChecker extends AbstractChecker
         }
 
         return true;
+    }
+
+    public function count($value, int $length): bool
+    {
+        if (!is_array($value) && !$value instanceof \Countable) {
+            return false;
+        }
+
+        return count($value) === $length;
+    }
+
+    public function shorter($value, int $length): bool
+    {
+        if (!is_array($value) && !$value instanceof \Countable) {
+            return false;
+        }
+
+        return count($value) <= $length;
+    }
+
+    public function longer($value, int $length): bool
+    {
+        if (!is_array($value) && !$value instanceof \Countable) {
+            return false;
+        }
+
+        return count($value) >= $length;
+    }
+
+    public function range($value, int $min, int $max): bool
+    {
+        if (!is_array($value) && !$value instanceof \Countable) {
+            return false;
+        }
+
+        $count = \count($value);
+
+        return $count >= $min && $count <= $max;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **High Impact Changes**
 - **Medium Impact Changes**
 - **Other Features**
+    - [spiral/validation] Add array::count, array::range, array::shorter and array::longer rules (#435)
 - **Bug Fixes**
 
 ## v2.8.0 - 2021-06-03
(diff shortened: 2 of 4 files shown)


src/Validation/src/Checker/ArrayChecker.php
```diff
@@ -31,4 +38,27 @@ public function of($value, $checker): bool
 
         return true;
     }
+
+    public function count($value, ?int $min = null, ?int $max = null, ?int $exact = null): bool
```
SerafimArts, line 42: Shouldn't `$min` be the default 0?

vladgorenkin: FIxed, thanks

SerafimArts:
Maybe it makes sense then to remove "exact" and change the "max" logic to `$max ??= $min`?

Then it will turn out:
```php
$checker->count($arr, 4); // Strictly count($arr) === 4
$checker->count($arr, 1, 42); // In range from 1 to 42
```

What do you think?

vladgorenkin:
And `$checker->count($arr, 1, null)` to only specify lower limit?

SerafimArts:
> And $checker->count($arr, 1, null) to only specify lower limit?

What about `$checker->count($arr, 1, INF)`?

vladgorenkin:
>  what about a more semantically consistent variant `$checker->count($arr, 1, INF)`?

\INF is problematic because it is special PHP value which cannot be exported without additional processing. 

```php
var_dump(json_encode(['array::count', 1, \INF]));
```
gives me `bool(false)`.

vladgorenkin:
It is also important to take into account that this function will mostly be used through validation DSL, where you can specify arguments names:
```php
['array::count', 'exact' => 4]
['array::count', 'min' => 2]
['array::count', 1, 5]
```

SerafimArts:
> \INF is problematic because it is special PHP value which cannot be exported without additional processing.

It is not entirely clear why these values should be converted into json?

vladgorenkin:
> > \INF is problematic because it is special PHP value which cannot be exported without additional processing.
> 
> It is not entirely clear why these values should be converted into json?

I see at least 2 possible use cases:
- to share validation rules with frontend
- to store in external storage. For instance, dynamic form with validation stored in DB

All existing rules are json-encodable, but this one isn't. I understand this might not be a priority for you, but I'm just sharing my concerns.

vladgorenkin:
We could also use `count($arr, 2, 2)` for exact case (as Symfony's count constraint does) but it looks ugly to me

SerafimArts:
> I see at least 2 possible use cases:

Yea, agree. I think this cases may well happen.

> We could also use count($arr, 2, 2) for exact case (as Symfony's count constraint does) but it looks ugly to me

Well, this option is better than 3 different arguments.

---

Or make two different validators? One with range and the other with a strict amount?

```
array::range($arr, 1); // 1+
array::range($arr, 1, 42); // 1 ... 42

// and

array::count($arr, 42); // 42
```

vladgorenkin:
I would say `range` is a bit unclear name but it's more consistent taking in mind existing `string::range` and `number::range`. Speaking of consistency: do we want to create something like `string::shorter/longer` but for arrays?

SerafimArts:
> I would say range is a bit unclear name

But it is also located under the "namespace" `array`, so there shouldn't be any problems with a misunderstanding of what this validator is doing. Isnt it?

vladgorenkin:
> > I would say range is a bit unclear name
> 
> But it is also located under the "namespace" `array`, so there shouldn't be any problems with a misunderstanding of what this validator is doing.

Well, my first association with `array::range(1, 42)` is a list of sequent number from 1 to 42 :) But ok I dont' mind. I'll update PR tomorrow

SerafimArts:
> Well, my first association with `array::range(1, 42)` is a list of sequent number from 1 to 42 :) 

What about `array::in_range(1, 42)`?

vladgorenkin:
> > Well, my first association with `array::range(1, 42)` is a list of sequent number from 1 to 42 :)
> 
> What about `array::in_range(1, 42)`?

`range` is better due to constistency.
I noticed a problem with `string::shorter/longer` and `number::lower/higher` -- they are actually `lower or equal` and `higher or equal`. What's worse is that validation messages do not reflect this either

SerafimArts (approved):

vladgorenkin:
Should I also check if number of elements in `shorter/longer` is equal to `$length` as it's done in other checkers? And should I adjust error messages in all checkers accordingly (unfortunately it's going to break translations for users)?

wolfy-j:
Yes it makes sense to adjust errors for all the checkers but mention that in PR and next release. Worse case it will fallback for English version.

vladgorenkin:
> Yes it makes sense to adjust errors for all the checkers but mention that in PR and next release. Worse case it will fallback for English version.

✔️

SerafimArts (approved):

PR #438, opened 2021-08-06
Generate annotated routes name when it is not specified explicitly

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️

This feature implementation is more of a dialog starter than a request for a change.

Speaking from my experience on a project with 250+ API-methods -- I never needed to reference a route name once. I understand  route names are useful for generating URLs and etc, but at the same time they can be an unnecessary burden. You have to make sure that names are unique (which I offen forget to check after copying controller actions). You also need to have sane naming convention and refactor route names with actions and etc.

Instead we could make specifying route names optional (at least for annotations) and when not specified generate them from route methods and pattern. In the current implementation route name is generated as follows:

```php
/** @Route('/api/news', methods={"POST", "PATCH"}) */ -> 'post,patch:/api/news'
```

What do you think of this idea? Does it make sense to you?

### DIFF
--- a/src/AnnotatedRoutes/src/Annotation/Route.php
+++ b/src/AnnotatedRoutes/src/Annotation/Route.php
@@ -28,10 +28,10 @@ final class Route
     public $route;
 
     /**
-     * @Attribute(name="name", type="string", required=true)
+     * @Attribute(name="name", type="string")
      * @var string
      */
-    public $name;
+    public $name = null;
 
     /**
      * @Attribute(name="verbs", type="mixed", required=true)
--- a/src/AnnotatedRoutes/src/RouteLocator.php
+++ b/src/AnnotatedRoutes/src/RouteLocator.php
@@ -42,8 +42,9 @@ final class RouteLocator
         foreach ($routes as $match) {
             /** @var RouteAnnotation $route */
             $route = $match->getAnnotation();
+            $routeName = $route->name ?? $this->generateName($route);
 
-            $result[$route->name] = [
+            $result[$routeName] = [
                 'pattern'    => $route->route,
                 'controller' => $match->getClass()->getName(),
                 'action'     => $match->getMethod()->getName(),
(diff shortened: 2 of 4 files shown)


SerafimArts:
The name of the route, in my opinion, should initially be optional. The fact that it will be generated now, I think the best option is not to break compatibility.

vladgorenkin:
> The name of the route, in my opinion, should initially be optional. The fact that it will be generated now, I think the best option is not to break compatibility.

Yes, names autogeneration is the easiest solution I could see to not break BC and leave ability to reference routes. Is `post,patch:/api/news` name format good enough? If yes I'd like to know what else is needed for this PR to be merged.

Thanks

src/AnnotatedRoutes/src/Annotation/Route.php
```diff
@@ -28,10 +28,10 @@ final class Route
     public $route;
 
     /**
-     * @Attribute(name="name", type="string", required=true)
+     * @Attribute(name="name", type="string")
      * @var string
```
SerafimArts, line 32: `@var string|null`

SerafimArts (approved):
LGTM 👍🏻

src/AnnotatedRoutes/src/Annotation/Route.php
vladgorenkin: Fixed

vladgorenkin: Ah, yes, there is a problem with union types and doctrine annotations:
```
 [Type Error] Attribute "name" of @Route declared on method Spiral\Tests\Router\App\Controller\HomeController::index() expects a(n) string|null, but got string. 
```
I guess I'll have to rollback this fix

SerafimArts: @vladgorenkin yep =(

PR #439, opened 2021-08-06
Use spiral/attribute for looking up metadata across project

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ✔️ if user has ReaderInterface that does not read annotations
| New feature?  | ✔️


SerafimArts:
wow!

src/Annotations/src/AnnotationLocator.php
```diff
@@ -32,12 +33,12 @@ final class AnnotationLocator implements SingletonInterface
...
+     * @param ClassesInterface     $classLocator
+     * @param ReaderInterface|null $reader
      *
      * @throws \Doctrine\Common\Annotations\AnnotationException
      */
-    public function __construct(ClassesInterface $classLocator, AnnotationReader $reader = null)
+    public function __construct(ClassesInterface $classLocator, ReaderInterface $reader = null)
     {
         $this->classLocator = $classLocator;
         $this->reader = $reader ?? new AnnotationReader();
```
SerafimArts, line 44: In this case, I would replace the `AnnotationReader` with a `SelectiveReader` one or `(new \Spiral\Attributes\Factory())->create()`, which uses both attributes and annotations.

vladgorenkin: Fixed

vladgorenkin:
As per telegram discussion I've added attribute declarations for following annotations: `@Route`, `@Guarded`/`@GuardNamespace`, `@Prototyped`, `@Pipeline` and `@DataGrid`. Corresponding readers and tests were updated as well

vladgorenkin:
Tests with `prefer-lowest` are failing because spiral/attributes:2.8 is needed for `NamedArgumentConstructor` to work, but currently root composer.json is only aware of 2.7 version which is a dependency of cycle/annotated. I'm not experienced with monorepos so I'd like to hear your input before making actions

SerafimArts:
Older (1.11) versions of `doctrine/annotations` use an interface instead of attributes and annotations: https://github.com/doctrine/annotations/blob/1.11.2/lib/Doctrine/Common/Annotations/NamedArgumentConstructorAnnotation.php 

Therefore, instead of `#[NamedArgumentConstructor]` (that has been added in 1.12: https://github.com/doctrine/annotations/blob/1.12.0/lib/Doctrine/Common/Annotations/Annotation/NamedArgumentConstructor.php), you should use `implements \Spiral\Attributes\NamedArgumentConstructorAttribut` for named constructor arguments.

src/AnnotatedRoutes/composer.json
```diff
@@ -16,8 +16,9 @@
     ],
     "require": {
         "php": ">=7.2",
-        "spiral/router": "^2.9",
-        "spiral/annotations": "^2.9"
+        "spiral/annotations": "^2.9",
+        "spiral/attributes": "^2.8",
```
SerafimArts, line 20: `"spiral/attributes": "^2.8"` -> `"spiral/attributes": "^2.9"`

src/Annotations/composer.json
```diff
@@ -10,8 +10,8 @@
     },
     "require": {
         "php": ">=7.2",
-        "spiral/tokenizer": "^2.9",
-        "doctrine/annotations": "^1.11"
+        "spiral/attributes": "^2.8",
```
SerafimArts, line 13: `"spiral/attributes": "^2.8"` -> `"spiral/attributes": "^2.9"`

SerafimArts (approved):

Issue #425, opened 2021-06-25
Wrong checker alias in ValidationBootloader


Eugentis:
In [ValidationBootloader.php](https://github.com/spiral/framework/blob/master/src/Framework/Bootloader/Security/ValidationBootloader.php) you have alias like `'required'   => 'type::notEmpty',` at line 90

This alias is not correct by system logic. Required and not empty are not equals.

required means that param should exists but it can be empty string or 0 as integer.

It can be used alias to `type::notNull` instead.

Eugentis:
@SerafimArts @wolfy-j 
Do you have any objections for provided MR?

roxblnfk:
I have some doubts about the issue.

If a string field is configured as `required', does that mean it must be filled?
If so then an empty string must not satisfy this rule.

At the same time, the PHP `empty()` function also doesn't work well because the value `"0"` look like empty value for PHP.

I think we should only fix the case with the `"0"` value.

roxblnfk:
> I think we should only fix the case with the "0" value.

And i see this case also covered

https://github.com/spiral/framework/blob/5dff8222f6ab4aae1afd02467afd9371cedb8788/src/Validation/src/Checker/Traits/NotEmptyTrait.php#L28-L30

msmakouz:
The current version of the `required` is similar in behavior to the `required` in Laravel:
https://laravel.com/docs/9.x/validation#rule-required

> The field under validation must be present in the input data and not empty. A field is considered "empty" if one of the following conditions are true:

    The value is null.
    The value is an empty string.
    The value is an empty array or empty Countable object.
    The value is an uploaded file with no path.

 And `NotBlank` in Symfony:
https://symfony.com/doc/current/reference/constraints/NotBlank.html

> Validates that a value is not blank - meaning not equal to a blank string, a blank array, false or null (null behavior is configurable).

We don't think we should make `required` an alias to `notNull`.

PR #441, opened 2021-08-15
[bugfix]: issue-425 - fix request validator require to not null

Eugentis:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ✔️
| New feature?  | ❌ 
| Issues        | #425

Required field can be empty but not null

This change can brokes validation if value will be empty but not null


butschster:
This PR will break a lot of projects. It will be redesigned in future versions

PR #444, opened 2021-08-23
Allow injectors to inject interfaces

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌

This PR allows to use injectors to inject with interface:

```php
//
// Bootloader
//
private pools = [
    'pool' => new ArrayAdapter(),
];
function boot(Container $container) { 
    $container->bindInjector(CacheItemPoolInterface::class, self::class)
}

function createInjection(\ReflectionClass $class, string $context = null) {
    return $this->pools[$context];
}
//...
// Code
//
public withClass(ArrayAdapter $pool) { // ArrayAdapter implements CachePoolItemInterface, injection works
}

public withInterface(CacheItemPoolInterface $pool) { // Currently container throws 'Undefined class or binding '%s'', because class_exists() returns false for CacheItemPoolInterface
}
```

### DIFF
--- a/src/Core/src/Container.php
+++ b/src/Core/src/Container.php
@@ -68,13 +68,6 @@ final class Container implements
     private $injectors = [];
 
     /**
-     * Contains names of all classes which were checked for the available injectors.
-     *
-     * @var array
-     */
-    private $injectorsCache = [];
-
-    /**
      * Container constructor.
      */
     public function __construct()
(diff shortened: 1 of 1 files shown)


vladgorenkin:
Bumping. Would you please take a look at this? I'd love to see this change in 2.9. Thanks

Issue #456, opened 2021-09-28
Attributes edge case: argument indices in wrong order


donquixote:
### Description

Numeric argument indices in wrong order behave differently in NamedArgumentsInstantiator vs native attributes. 

### How To Reproduce

Insert in `NamedArgumentsInstantiatorTestCase::testNamedConstructorInstantiatable()`.

```php
        $object = $this->new(NamedArgumentConstructorFixture::class, [
            1 => 'one',
            0 => 'zero',
        ]);

        if (PHP_VERSION_ID < 80000) {
            $this->assertSame('zero', $object->a);
            $this->assertSame('one', $object->b);
        }
        else {
            $this->assertSame('one', $object->a);
            $this->assertSame('zero', $object->b);
        }
        
        $this->assertSame(null, $object->c);
```
The test will pass.
It shows that `NamedArgumentsInstantiator::resolveParameters()` does not fully replicate PHP 8 named parameters behavior.

This is an edge case, but PHP 8 does not produce an error if the indices are inversed.

donquixote:
I have a pull request ready for `spiral/attributes`. It also adds new tests.
But now I wonder if instead I should make a pull request here in `spiral/framework`?
How can I bring my branch from the subtree split back into the common repo?

SerafimArts:
Yep. All PRs should be directed to this repository (in this case, branch `2.8`).

I think the easiest way is to just copy the files from your repository into the `src/Attributes/*` directory and commit the changes.

donquixote:
> I think the easiest way is to just copy the files from your repository into the src/Attributes/* directory and commit the changes.

Is there no way to do it on git level?
I have multiple commits that I would like to keep in sequence.
But ok, will do it the hard way this time :)

SerafimArts:
> Is there no way to do it on git level?

IDK. I have never encountered such a thing ¯\\\_(ツ)_/¯

donquixote:
I did the following:
- Add my local attributes repo as a remote. Fetch.
- Cherry-pick the 3 commits from the feature branch in the attributes repo. Git uses heuristics to determine the file and directory where changes should be applied, but these heuristics are limited.
- Use `git rebase -ir HEAD^^^´, move files that were created in the wrong place, and amend the respective commits.

PR #459, opened 2021-09-28
Issue #456: Align arguments resolving in NamedArgumentsInstantiator with native PHP 8 behavior.

donquixote:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️ or TBD, see changes in tests.
| Breaks BC?    | ❌ or TBD, see changes in tests.
| New feature?  | ❌ or TBD, see changes in tests.
| Issues        | #456 
| Docs PR       | (none)

The arguments resolving in `NamedArgumentsInstantiator` has some edge cases where it diverges from native behavior in PHP 8. This PR adds more test cases and attempts to align the behavior more with PHP 8.

(I might still do some tweaking with force-push)


SerafimArts:
Master branch is the `2.9.x-dev` version. Are you sure these fixes should be reserved for the next minor release instead of `2.8.x` patch?

donquixote:
I am splitting the big function, but I am not sure if that's really so helpful.
It is nice if local variables have a limited scope. But does the split really create meaningful separation?

donquixote:
@SerafimArts I recommend to keep the commit history of this PR.
Each commit is meant to have passing tests, so we see the effect of before/after.
Only the last commit (splitting the big method) we could squash with the previous one, but we should first fine-tune.

donquixote:
> Master branch is the 2.9.x-dev version. Are you sure these fixes should be reserved for the next minor release instead of 2.8.x patch?

Ah. Sure, I can use 2.8.x as a base instead.

donquixote:
Base changed.

donquixote:
I forgot to add a test for the variadic parameter. Going to do it now.

SerafimArts:
@donquixote It's okay, dont worry. I haven't started (as you might have noticed) the review yet.

The current code for the `appendNamedArgs` function looks pretty impressive and complex, so to understand whether it can be simplified, I need to investigate its work. But since I have other tasks in my work now, I may be able to do a review either by the end or next week.

donquixote:
Improved test coverage and support for various edge cases.

butschster (approved):

roxblnfk (approved):

PR #466, opened 2021-10-20
Exit cycle:migrate with failure code when outstanding migrations are found

vladgorenkin:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌


butschster:
Will be moved to the cycle-bridge package

PR #469, opened 2021-11-02
PlainRenderer fix for multiply `Accept` header

gam6itko:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ✔️ 
| Issues        | #... 

Client send `Accept` header.

I have this lines in my Roadrunner config.
```yaml
headers:
    request:
        "Accept": "application/json"
```

If an error occurs `PlainRenderer` could not read the header correctly and sends response not in JSON format

### DIFF
--- a/src/Framework/Http/ErrorHandler/PlainRenderer.php
+++ b/src/Framework/Http/ErrorHandler/PlainRenderer.php
@@ -14,6 +14,7 @@ namespace Spiral\Http\ErrorHandler;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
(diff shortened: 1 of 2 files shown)


roxblnfk:
The fix looks correct but in case when `application/json` has lower priority then `html` a error will be rendered in json
```
Accept: text/html, application/json;q=0.9, */*;q=0.8
```

I think it is rare case but can you add test for this?

gam6itko:
@roxblnfk ok, I'll do this

gam6itko:
@roxblnfk ready for review

roxblnfk (approved):

wolfy-j (approved):

butschster:
Thank you @gam6itko for contribution! We are looking forward new PRs from you :)

PR #471, opened 2021-11-10
[Rector] Start Add Rector for QA

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

This PR apply Rector as part of QA check, start with:

- `LevelSetList::UP_TO_PHP_72` to apply syntax support until php 7.2 per-composer.json
- Only apply to `src/*/src` only (for now)
- Use pinned version `0.12.3` on purpose as it need to sync phpstan and nikic/php-parser, ensure only run when working ok.
- Skip `CountOnNullRector` on purpose as it got too detailed on some cases.
- Using `--dry-run` to show the diff output if there is expected change by rector, to proces, we can run:

```bash
vendor/bin/rector
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
         run: composer require --dev roave/security-advisories:dev-latest
 
   static-analysis:
-    name: Psalm
+    name: Psalm and Rector
     runs-on: ${{ matrix.os }}
 
     strategy:
@@ -100,7 +100,9 @@ jobs:
 
       # Execution
       - name: Static Analysis
-        run: vendor/bin/psalm --no-cache
+        run: |
+          vendor/bin/psalm --no-cache
+          vendor/bin/rector --dry-run
 
 
   coding-standards:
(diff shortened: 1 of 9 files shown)


butschster:
@samsonasik @wolfy-j Do we need to skip `CountOnNullRector`? As I remember we want to use >= php8 and `count(null)` will cause fatal error according to this example https://3v4l.org/Bndc9. And if we have a code like `count(null)` then we need to get rid of it.

samsonasik:
@butschster I applied the `CountOnNullRector` at https://github.com/spiral/framework/pull/471/commits/47c3628801f10c4ce53fccf216efab79369fb39a

If that's too detailed, I can revert to skip it.

samsonasik:
@butschster improvement to rector is coming https://github.com/rectorphp/rector-src/pull/1208 for `CountOnNullRector`

butschster:
> @butschster improvement to rector is coming [rectorphp/rector-src#1208](https://github.com/rectorphp/rector-src/pull/1208) for `CountOnNullRector`

You are right `CountOnNullRector` adds more noise. I think we can skip it now, because there are only couple of places where `count` method used and check them later.

samsonasik:
@butschster done

butschster (approved):

butschster:
Thank you @samsonasik for contribution!

PR #472, opened 2021-11-11
[Rector] Apply SetList::DEAD_CODE

samsonasik:
| Q             | A
| ------------- | ---
| QA?       | ✔️

Apply Rector's `SetList::DEAD_CODE` Setlist to remove unnecessary code.

Mostly applied changes:

- unnecessary docblock (param, return, when exactly same with no description)
- unnecessary default value on property when always replaced in __construct.
- remove default null on property as non-typed yet.
- remove unnecessary cast as already typed.
- remove parameter in last when never used in class method stmts
- remove unnecessary assign.

Complete lists are at https://github.com/rectorphp/rector-src/blob/6a8d8233b963f3361383ce7a76a343be16837bff/config/set/dead-code.php


src/Console/src/Sequence/CallableSequence.php
```diff
@@ -23,23 +23,15 @@ final class CallableSequence extends AbstractSequence
...
-
     /**
      * @param callable $function
-     * @param array    $parameters
-     * @param string   $header
-     * @param string   $footer
      */
     public function __construct(
         $function,
-        array $parameters = [],
```
samsonasik, line 37: The private `$parameters` property actually never used, but remove parameter in the middle seems make a bug, I will revert and skip `RemoveUnusedPrivatePropertyRector`

samsonasik: reverted and skipped `RemoveUnusedPrivatePropertyRector`

samsonasik:
I will look into the errors, possibly skip rector rules that cause it.

rector.php
```diff
@@ -24,6 +26,10 @@
 
         // buggy when remove private property, but filled by construct in the middle of parameter
         RemoveUnusedPrivatePropertyRector::class,
+
+        RemoveUnusedPrivateMethodRector::class => [
+            ConfigurationBootloader::class,
+        ],
```
samsonasik, line 32: keep private method `ConfigurationBootloader::configManager()` on purpose as it used in singleton:

https://github.com/spiral/framework/blob/cc6cb37254e8185561f12c9e6d1c0ade824ff9a3/src/Boot/src/Bootloader/ConfigurationBootloader.php#L33

rector.php
```diff
@@ -15,7 +20,18 @@
...
         CountOnNullRector::class,
+
+        // for PHP 8
+        RemoveUnusedPromotedPropertyRector::class,
+
+        // buggy when remove private property, but filled by construct in the middle of parameter
+        RemoveUnusedPrivatePropertyRector::class,
+
+        RemoveUnusedPrivateMethodRector::class => [
+            __DIR__ . '/src/Boot/src/Bootloader/ConfigurationBootloader.php',
```
samsonasik, line 31: keep private method `ConfigurationBootloader::configManager()` on purpose as it used in singleton:

https://github.com/spiral/framework/blob/cc6cb37254e8185561f12c9e6d1c0ade824ff9a3/src/Boot/src/Bootloader/ConfigurationBootloader.php#L33

src/Dumper/src/System.php
```diff
@@ -60,18 +58,7 @@ function_exists('sapi_windows_vt100_support')
...
-
-            if (\function_exists('posix_isatty')) {
-                return @posix_isatty($stream);
-            }
-
-            $stat = @fstat($stream);
-
-            // Check if formatted mode is S_IFCHR
-            return $stat ? 0020000 === ($stat['mode'] & 0170000) : false;
+            return @stream_isatty($stream);
```
samsonasik, line 61: `stream_isatty()` exists in php 7.2+, so no need to use function_exists()

src/Core/src/Container.php
```diff
@@ -503,7 +473,7 @@ private function createInstance(string $class, array $parameters, string $contex
 
             $instance = null;
             try {
-                /** @var InjectorInterface $injectorInstance */
+                /** @var InjectorInterface|mixed $injectorInstance */
```
samsonasik, line 476: added union `/** @var InjectorInterface|mixed $injectorInstance */` to make `if (!$injectorInstance instanceof InjectorInterface) {` next valid.

src/Http/src/Header/AcceptHeader.php
```diff
@@ -128,6 +112,7 @@ private function addItem($item): void
             $item = AcceptHeaderItem::fromString((string)$item);
         }
 
+        /** @var AcceptHeaderItem|array $item  */
```
samsonasik, line 115: added union `/** @var AcceptHeaderItem|array $item  */` to make `if (!$item instanceof AcceptHeaderItem) {` next valid.

samsonasik:
@butschster all green 🎉

rector.php
```diff
@@ -15,7 +20,18 @@
 
     $parameters->set(Option::SKIP, [
         CountOnNullRector::class,
+
+        // for PHP 8
+        RemoveUnusedPromotedPropertyRector::class,
+
+        // buggy when remove private property, but filled by construct in the middle of parameter
+        RemoveUnusedPrivatePropertyRector::class,
```
samsonasik, line 28: improvements are coming to rector:

- https://github.com/rectorphp/rector-src/pull/1212
- https://github.com/rectorphp/rector-src/pull/1214

src/Dumper/src/System.php
samsonasik: @butschster what "." means, missclick submit comment?

src/AuthHttp/src/Middleware/Firewall/AbstractFirewall.php
```diff
@@ -23,14 +23,8 @@
...
 abstract class AbstractFirewall implements MiddlewareInterface
 {
-    /**
-     * @param Request                 $request
-     * @param RequestHandlerInterface $handler
-     * @return Response
-     */
     public function process(Request $request, RequestHandlerInterface $handler): Response
     {
-        /** @var AuthContextInterface $context */
```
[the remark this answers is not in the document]
samsonasik: there was no `$context` variable, the variable doc should be changed to `$authContext`, I updated to `$authContext`

https://github.com/spiral/framework/pull/472/commits/9eb873ef7d5ce288da31689986895e3271f8bde1

src/Csrf/src/Middleware/CsrfFirewall.php
```diff
@@ -85,14 +78,11 @@ protected function isRequired(Request $request): bool
...
     /**
      * Fetch token from request.
-     *
-     * @param Request $request
-     * @return string
      */
     protected function fetchToken(Request $request): string
     {
         if ($request->hasHeader(self::HEADER)) {
-            return (string)$request->getHeaderLine(self::HEADER);
```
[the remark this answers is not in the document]
samsonasik: The `getHeaderLine()` always return string per-psr `MessageInterface`, also already checked with `hasHeader()` previously.

`laminas-diactoros` return string for it, ` guzzle/psr7` return string as well, I think if there is a case that it is not string, then that's the implementor bug.

src/DataGrid/src/Specification/Filter/ValueBetween.php
```diff
@@ -161,23 +148,17 @@ private function invalidExpressionType($value): string
...
             : new Gt($this->value[1], $this->expression);
     }
 
-    /**
-     * @return FilterInterface
-     */
     private function toFilter(): FilterInterface
     {
         return $this->includeTo
-            ? new Lte((string)$this->value[0], $this->expression)
```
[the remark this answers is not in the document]
samsonasik: It based on `@var string[]` in the property definition

https://github.com/spiral/framework/blob/cc6cb37254e8185561f12c9e6d1c0ade824ff9a3/src/DataGrid/src/Specification/Filter/ValueBetween.php#L23-L24

ref https://phpstan.org/r/0bb07df6-b0e3-4a1e-928d-9157e671fe73

If there is possible not string array data, then value property need to be changed to something like:

```php
/** @var string[]|int[] */ 
 private $value; 
```

https://phpstan.org/r/378d473f-cbeb-422d-82b3-9c695989e777

src/Attributes/src/polyfill.php
```diff
@@ -118,17 +118,6 @@ final class Attribute
              * @var int-mask-of<Attribute::TARGET_*>
              */
             public $flags;
-
-            /**
-             * @param int-mask-of<Attribute::TARGET_*> $flags A value in the
-             *        form of a bitmask indicating the places where attributes
-             *        can be defined.
-             */
-            public function __construct(
```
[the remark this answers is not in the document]
roxblnfk: Now this code is dead. Mb is it mistake?

butschster:
@samsonasik hi! Sorry for long response. We didn't forget about your your PR, this week I'll try to check it and merge! Thank you!

src/Router/src/Target/Namespaced.php
```diff
@@ -30,16 +30,12 @@ final class Namespaced extends AbstractTarget
     private $inflector;
 
     /**
-     * @param string $namespace
-     * @param string $postfix
-     * @param int    $options
      * @param string $defaultAction
```
butschster: I think. we have to get rid of this line

src/Router/src/Target/Namespaced.php
```diff
@@ -30,16 +30,12 @@ final class Namespaced extends AbstractTarget
...
      */
     public function __construct(
         string $namespace,
         string $postfix = 'Controller',
-        int $options = 0,
-        string $defaultAction = 'index'
```
butschster: @roxblnfk What do you think about this argument?

src/Router/src/Target/Namespaced.php
samsonasik: removed https://github.com/spiral/framework/pull/472/commits/6818a95bda8decc5f5579873eb9c98cd2ebfcb83

src/Router/src/Target/Namespaced.php
samsonasik: `__construct` is a special method which doesn't need to follow parent count parameter, and the `defaultAction` parameter is never used in the class method stmts, so I think removal is safe.

roxblnfk: Looks OK

PR #489, opened 2022-01-22
[Rector] Update to rector 0.12.15, apply parallel, and RemoveUnusedPrivatePropertyRector

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

- updated rector dependency to 0.12.15
- apply parallel process to speed up check
- apply RemoveUnusedPrivatePropertyRector
- Enable allow-plugins for composer 2.2 support.


samsonasik:
All green 🎉

rector.php
```diff
@@ -18,14 +17,17 @@
...
         CountOnNullRector::class,
 
         // for PHP 8
         RemoveUnusedPromotedPropertyRector::class,
 
-        // buggy when remove private property, but filled by construct in the middle of parameter
-        RemoveUnusedPrivatePropertyRector::class,
+        RemoveUnusedPrivatePropertyRector::class => [
+            // buggy when used in middle of parameter along with RemoveUnusedConstructorParamRector that may cause BC break
+            __DIR__ . '/src/Console/src/Sequence/CallableSequence.php',
```
samsonasik, line 29: this skipped part resolved at https://github.com/rectorphp/rector-src/pull/1714 , can be cleaned up on next rector release.

samsonasik:
updated to use rector 0.12.4 and re-run rector so option skip now cleaned up.

src/Console/src/Sequence/CallableSequence.php
```diff
@@ -36,7 +33,6 @@ public function __construct(
         string $footer = ''
     ) {
         $this->function = $function;
-        $this->parameters = $parameters;
```
samsonasik, line 39: only remove property and its assignment in constructor for unused private property without removing its middle `__construct` parameter that fill it to avoid BC break.

samsonasik:
All green 🎉

butschster:
Hi @samsonasik! Thank you very much! You are the best! We are going to start working on spiral framework 3.0 and it would be great to prepare rector for php 8.0!

samsonasik:
@butschster yes 👍 , just let me know there is a branch 3.0 to give php 8.0 rector rules to apply

samsonasik:
updated to use rector 0.12.15

samsonasik:
All green 🎉

butschster (approved):

butschster:
> @butschster yes 👍 , just let me know there is a branch 3.0 to give php 8.0 rector rules to apply

I think, you can start from 2.9 branch

samsonasik:
2.9 branch still using php `>=7.2`: 

https://github.com/spiral/framework/blob/3dca6451950fab790b880e2fd0b7ec535bcb047d/composer.json#L21

is it needs to be updated to `>=8.0`? For safer in the future, I think it is better to use:

```javascript
"php": "~8.0.0 || ~8.1.0"
```

since php is not semver.

butschster:
> 2.9 branch still using php `>=7.2`:
> 
> https://github.com/spiral/framework/blob/3dca6451950fab790b880e2fd0b7ec535bcb047d/composer.json#L21
> 
> is it needs to be updated to `>=8.0`? For safer in the future, I think it is better to use:
> 
> ```js
> "php": "~8.0.0 || ~8.1.0"
> ```
> 
> since php is not semver.

+

PR #503, opened 2022-01-31
Adds HasLocaleInterface 

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️

Adds HasLocaleInterface for cases when we need to change default locale

### DIFF
--- a/src/Translator/src/HasLocaleInterface.php
+++ b/src/Translator/src/HasLocaleInterface.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


butschster:
It can be replaced with [https://github.com/symfony/translation-contracts/blob/main/LocaleAwareInterface.php](LocaleAwareInterface)

butschster:
https://github.com/symfony/translation-contracts/blob/main/LocaleAwareInterface.php will be used instead

Issue #550, opened 2022-02-08
Rename mailer `pipeline` config key to `queue`


butschster:
And keep support for old key

PR #556, opened 2022-02-08
Added config param `queue` to the SendIt

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | #550

### DIFF
--- a/src/SendIt/src/Config/MailerConfig.php
+++ b/src/SendIt/src/Config/MailerConfig.php
@@ -21,6 +21,7 @@ final class MailerConfig extends InjectableConfig
     protected $config = [
         'dsn' => '',
         'from' => '',
+        'queue' => '',
         'pipeline' => '',
         'queueConnection' => null,
     ];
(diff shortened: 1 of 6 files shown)


src/SendIt/src/Config/MailerConfig.php
```diff
@@ -21,6 +21,7 @@ final class MailerConfig extends InjectableConfig
     protected $config = [
         'dsn' => '',
         'from' => '',
+        'queue' => '',
```
butschster, line 24: it should be `null` as `pipeline`

src/SendIt/src/Config/MailerConfig.php
```diff
@@ -35,9 +36,17 @@ public function getFromAddress(): string
...
+     */
     public function getQueuePipeline(): string
     {
-        return $this->config['pipeline'];
+        return $this->getQueue();
+    }
+
+    public function getQueue(): string
+    {
+        return $this->config['queue'] ?? $this->config['pipeline'];
```
butschster, line 49: `return $this->config['queue'] ?? $this->config['pipeline'] ?? null;`

PR #559, opened 2022-02-09
class annotations for traits

aquaminer:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️ 

The following example, which is incapable of functioning without this feature:
```php
#[Behavior\CreatedAt]
#[Behavior\UpdatedAt]
trait TsTrait
{
    #[Cycle\Column(type: 'datetime')]
    private DateTimeImmutable $createdAt;

    #[Cycle\Column(type: 'datetime', nullable: true)]
    private ?DateTimeImmutable $updatedAt = null;
}

#[Cycle\Entity]
class Entity {
    use TsTrait;
}
```


roxblnfk:
I like it

PR #573, opened 2022-02-17
[PHPUnit] Fix PHPUnit notice Cannot acquire reference to $GLOBALS on lowest dep

samsonasik:
| Q             | A
| ------------- | ---
| QA?       | ✔️

<!-- Please, replace this notice by a short description of your feature/bugfix.

Ref https://stackoverflow.com/questions/70455395/phpunit-cannot-acquire-reference-to-globals/70974793#70974793

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -205,7 +205,6 @@ jobs:
           max_attempts: 5
           command: composer require phpunit/phpunit:^9.5 --prefer-dist --no-interaction --no-progress
       - name: Install Dependencies
-        if: ${{ matrix.php < '8.1' }}
         uses: nick-invision/retry@v1
         with:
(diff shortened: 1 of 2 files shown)


samsonasik:
It seems lowest of phpunit 9.5.0 cause this error : 

```bash
PHP Fatal error:  During inheritance of IteratorAggregate: Uncaught Return type of Symfony\Component\Finder\Finder::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```

Needs to find when the lowest compatible phpunit for `#[\ReturnTypeWillChange]` addition.

samsonasik:
It seems various packages need to be updated for minimum version requirement for --prefer-lowest

PR #579, opened 2022-02-21
Fix installer downloader url

rohithmeethal:
Currently github has urls like `*download/2.9.0/*` Instead of  `*download/v2.9.0/*` (v prefix for version not required).
This was causing installation failure when running 
composer create-project spiral/app spiral-test

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | #... 

### DIFF
--- a/bin/spiral
+++ b/bin/spiral
@@ -127,7 +127,7 @@ class RRHelper
             $ext = '.tar.gz';
         }
 
-        return 'https://github.com/spiral/framework/releases/download/v'
+        return 'https://github.com/spiral/framework/releases/download/'
             . static::getVersion() . '/' . self::getSignature()
             . $ext;
     }


butschster:
Hi @rohithmeethal! Thank you for contribution!!!

With this PR you will break download links for previous releases.

We will support RR 1.x  until Spiral Framework 3.x and it it would be great if you write something like this:

```php
$version = static::getVersion();
if(version_compare($version, '2.9.0', '<')) {
     $version = 'v'.$version;
}

return 'https://github.com/spiral/framework/releases/download/'
            . $version . '/' . self::getSignature()
            . $ext;
```

PR #599, opened 2022-03-17
Adds an ability to disable overwriting env variables for Environment.php

### LINKED ISSUE #509: Fix order for loading Environment variables
We can add `DISABLE_ENV_OVERWRITING` variable to prevent overwriting and check it before creating Environment object

butschster:
fixes #509

| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️
| Issues        | #509

### DIFF
--- a/src/Boot/src/Environment.php
+++ b/src/Boot/src/Environment.php
@@ -23,15 +23,14 @@ final class Environment implements EnvironmentInterface
         'empty'   => '',
     ];
 
-    /** @var string|null */
-    private $id;
+    private ?string $id = null;
+    private array $values;
+    private bool $overwrite;
 
-    /** @var array */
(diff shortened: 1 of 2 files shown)


src/Boot/src/Environment.php
```diff
@@ -51,6 +50,10 @@ public function getID(): string
      */
     public function set(string $name, $value): void
     {
+        if (isset($this->values[$name]) && !$this->overwrite) {
```
roxblnfk, line 53:
```suggestion
        if (\array_key_exists($name, $this->values) && !$this->overwrite) {
```
It will work with predefined `null` value

roxblnfk (approved):

PR #615, opened 2022-03-24
Updating Config code with PHP 8.1 features

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ✔️
| New feature?  | ❌ 
| Issues        | #512


src/Config/src/Patch/Group.php
```diff
@@ -15,9 +8,11 @@
 
 final class Group implements PatchInterface
 {
-    /** @var array|PatchInterface[] */
-    private $patches = [];
+    private array $patches;
```
roxblnfk, line 11:
```suggestion
    /** @var PatchInterface[] */
    private array $patches;
```

src/Config/src/Loader/PhpLoader.php
```diff
@@ -20,23 +13,15 @@
...
-    /**
-     * @inheritdoc
-     */
     public function loadFile(string $section, string $filename): array
     {
         try {
-            return ContainerScope::runScope($this->container, function () use ($filename) {
-                return (require $filename);
-            });
+            return ContainerScope::runScope($this->container, fn () => require $filename);
```
roxblnfk, line 24:
```suggestion
            return ContainerScope::runScope($this->container, static fn () => require $filename);
```

PR #661, opened 2022-04-14
Adding chunkSize to the HTTP config

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

### DIFF
--- a/src/Framework/Bootloader/Http/HttpBootloader.php
+++ b/src/Framework/Bootloader/Http/HttpBootloader.php
@@ -103,6 +103,17 @@ final class HttpBootloader extends Bootloader implements SingletonInterface
         $this->config->modify('http', new Append('middleware', null, $middleware));
     }
 
+    public function createEmitter(HttpConfig $config): EmitterInterface
+    {
+        $emitter = new SapiEmitter();
+
+        if (($chunkSize = $config->getChunkSize()) !== null) {
+            $emitter->bufferSize = $chunkSize;
+        }
+
+        return $emitter;
+    }
+
     /**
      * @param HttpConfig               $config
      * @param Pipeline                 $pipeline
@@ -122,19 +133,5 @@ final class HttpBootloader extends Bootloader implements SingletonInterface
         $core->setHandler($handler);
 
         return $core;
-    }
-
-    /**
-     * @noRector RemoveUnusedPrivateMethodRector
-     */
-    private function createEmitter(HttpConfig $config): EmitterInterface
-    {
-        $emitter = new SapiEmitter();
-
-        if (($chunkSize = $config->getChunkSize()) !== null) {
-            $emitter->bufferSize = $chunkSize;
-        }
-
-        return $emitter;
     }
 }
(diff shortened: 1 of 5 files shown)


src/Http/src/Config/HttpConfig.php
```diff
@@ -51,4 +52,12 @@ public function getMiddleware(): array
     {
         return $this->config['middleware'] ?? $this->config['middlewares'];
     }
+
+    /**
+     * If chunkSize isn't provided - using default values
+     */
+    public function getChunkSize(): ?int
+    {
+        return !\is_null($this->config['chunkSize']) ? (int) $this->config['chunkSize'] : null;
```
roxblnfk, line 61: returned value should be positive int or null

src/Framework/Bootloader/Http/HttpBootloader.php
```diff
@@ -122,4 +123,18 @@ protected function httpCore(
 
         return $core;
     }
+
+    /**
+     * @noRector RemoveUnusedPrivateMethodRector
+     */
+    private function createEmitter(HttpConfig $config): EmitterInterface
```
roxblnfk, line 130: mb just do it public?

src/Http/src/Config/HttpConfig.php
msmakouz: Added a check that the value is greater than 0

src/Framework/Bootloader/Http/HttpBootloader.php
msmakouz: Done

src/Http/src/Config/HttpConfig.php
```diff
@@ -55,9 +55,15 @@ public function getMiddleware(): array
 
     /**
      * If chunkSize isn't provided - using default values
+     *
+     * @psalm-return positive-int
      */
     public function getChunkSize(): ?int
     {
-        return !\is_null($this->config['chunkSize']) ? (int) $this->config['chunkSize'] : null;
+        if (\is_null($this->config['chunkSize']) || (int) $this->config['chunkSize'] <= 0) {
```
roxblnfk, line 63:
```suggestion
        if (\is_null($this->config['chunkSize']) || (int) $this->config['chunkSize'] < 0) {
```

I propose to support `0` value. We need to update SapiEmitter also:
```php
// SapiEmitter::emitBody
if ($this->bufferSize === 0) {
    return (string) $body;
}
while (!$body->eof()) {
// ...
```
And need to add comment for the `SapiEmitter::$bufferSize` property about the `0` value that means disabled streamed response

msmakouz: Done.

roxblnfk (approved):

src/Http/src/Config/HttpConfig.php
```diff
@@ -51,4 +52,16 @@ public function getMiddleware(): array
...
+
+    /**
+     * If chunkSize isn't provided - using default values
+     */
+    public function getChunkSize(): ?int
+    {
+        if (\is_null($this->config['chunkSize']) || (int) $this->config['chunkSize'] < 0) {
+            return null;
+        }
+
```
butschster, line 64: Maybe do you replace `(int) $this->config['chunkSize'] < 0` with `assert((int) $this->config['chunkSize'] > 0, 'Chunk size should be greater than 0.')` ?

butschster (changes requested):

src/Http/src/Config/HttpConfig.php
roxblnfk: Chunk size can be equal 0

PR #668, opened 2022-04-21
Improves broadcasting guard.

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

### DIFF
--- a/src/Broadcasting/src/GuardInterface.php
+++ b/src/Broadcasting/src/GuardInterface.php
@@ -4,9 +4,15 @@ declare(strict_types=1);
 
 namespace Spiral\Broadcasting;
 
+use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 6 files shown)


src/Broadcasting/src/GuardInterface.php
```diff
@@ -4,9 +4,15 @@
...
 
 interface GuardInterface
 {
-    public function authorize(ServerRequestInterface $request): bool;
+    /**
+     * Authenticate a websocket connection request.
+     */
+    public function authorize(
+        ServerRequestInterface $request
+    ): ResponseInterface;
```
roxblnfk, line 17: Can we use DTO there? Like
success: bool
topics: string[]|string
response: ResponseInterface

roxblnfk (approved):
what about changelog?

PR #696, opened 2022-05-25
Improving version checker in MonorepoBuilder

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ❌


roxblnfk:
The same should be merged with 2.0 also

PR #698, opened 2022-05-26
Adding the ability to add input bags

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

### DIFF
--- a/src/Http/src/Request/InputManager.php
+++ b/src/Http/src/Request/InputManager.php
@@ -52,12 +52,12 @@ final class InputManager implements SingletonInterface
         'headers'    => [
             'class'  => HeadersBag::class,
             'source' => 'getHeaders',
-            'alias'  => 'header'
+            'alias'  => 'header',
         ],
         'data'       => [
             'class'  => InputBag::class,
             'source' => 'getParsedBody',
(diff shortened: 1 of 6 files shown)


src/Http/src/Request/InputManager.php
```diff
@@ -92,8 +107,11 @@ final class InputManager implements SingletonInterface
 
     public function __construct(
         /** @invisible */
-        private ContainerInterface $container
+        private readonly ContainerInterface $container,
+        /** @invisible */
+        HttpConfig $config = new HttpConfig()
     ) {
+        $this->bagAssociations = \array_merge($this->bagAssociations, $config->getInputBags());
```
butschster, line 114: Maybe we will get all aliases and create a new array with `alias` => `name`?

msmakouz: In this case, need to add a new method addInputBagAlias so that allow can add an alias. Now it can be done like this:
https://github.com/spiral-packages/symfony-validator/blob/master/src/Bootloader/ValidatorBootloader.php#L31

PR #713, opened 2022-06-15
Adds ApplicationInProduction confirmation class for console commands

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️

- [x] Code
- [x] Tests
- [ ] Docs

```php
use Spiral\Console\Confirmation\ApplicationInProduction;
final class MigrateCommand extends Command
{
    protected const NAME = 'db:migrate';
    protected const DESCRIPTION = '...';

    public function perform(ApplicationInProduction $confirmation): int
    {
        if (!$confirmation->confirmToProceed()) {
            return self::FAILURE;
        }
        
        // run migrations...
    }
}
```


src/Framework/Command/Encrypter/KeyCommand.php
```diff
@@ -40,8 +44,13 @@ public function perform(EncrypterFactory $enc, FilesInterface $files): int
             return self::SUCCESS;
         }
 
+        if (!$confirmation->confirmToProceed()) {
+            return self::FAILURE;
+        }
+
         if (!$files->exists($file)) {
-            $this->sprintf('<error>Unable to find `%s`</error>', $file);
+            $this->error(\sprintf('Unable to find `%s`', $file));
```
roxblnfk, line 52:
```suggestion
            $this->error(\sprintf('Unable to find `%s`.', $file));
```

roxblnfk (changes requested):

PR #736, opened 2022-07-17
[Rector] Update rector and re-run with LevelSetList::UP_TO_PHP_74 - take 2

samsonasik:
| Q             | A
| ------------- | ---
| QA?       | ✔️

This PR do the followings:

- Update rector dependency to latest 0.13.8
- Update LevelSetList to `LevelSetList::UP_TO_PHP_74` as current composer.json require php 7.4, and run it.
- Set auto import option to make typed property use its name instead of fqcn when possible, with exclude short class name.

Note: For update to typed property, to avoid BC break:

- On non-final class, it only update private property only when possible.
- On final class, it allow update protected property that doesn't has parent property usage.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -141,7 +141,7 @@
         "phpunit/phpunit": "^8.5|^9.5",
         "nyholm/psr7": "^1.5.0",
         "ramsey/uuid": "^4.2",
-        "rector/rector": "0.12.15",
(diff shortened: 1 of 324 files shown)


samsonasik:
@roxblnfk it is ready for review.

roxblnfk (approved):

PR #737, opened 2022-07-19
Introduce self-rendering filters

kafkiansky:
| Q                    | A
| ------------- | ---
| Bugfix?           | ❌
| Breaks BC?    | ✔️ 
| New feature? | ✔️

Allows to override the error rendering for a specific filter. 

**Example of usage**

The renderer:

```php
<?php
declare(strict_types=1);

namespace App;

use Spiral\Filters\FilterInterface;
use Spiral\Filters\RenderErrorsInterface;

/**
 * @template-implements RenderErrorsInterface<FilterInterface>
 */
final class SomeErrorsRenderer implements RenderErrorsInterface
{
    /**
     * {@inheritdoc}
     */
    public function render(FilterInterface $filter)
    {
        return [
            'success' => false,
            'errors' => $filter->getErrors(),
        ];
    }
}

```

The filter:

```php
<?php

declare(strict_types=1);

namespace App;

use Spiral\Filters\Filter;
use Spiral\Filters\RenderWith;

#[RenderWith(SomeErrorsRenderer::class)]
final class FilterWithErrorsRenderer extends Filter
{
    public const SCHEMA = [
        'id' => 'query:id'
    ];

    public const VALIDATES = [
        'id' => [
            ['notEmpty', 'err' => '[[ID is not valid.]]']
        ]
    ];
}
```


butschster:
Hi, @kafkiansky!

Thank you for the PR! Could you provide an example of using a new feature and some tests?

kafkiansky:
Hi, @butschster. I wanted to add tests, but I couldn't find where to put them. Perhaps there are already tests for `FilterInterceptor`?

And the example of use is simple. We globally use `FilterInterceptor` in our project, which validates all filters before they hit the controller, which is very convenient. However, in some cases we need a specific response for validation errors, not the one used throughout the project: for example, one format is used for internal API, but for external systems we need to return a different format for validation errors. Using the `ShouldRenderErrors` interface will allow us to influence the format of errors in specific cases.

butschster:
@kafkiansky It would be great to have something like in this PR https://github.com/spiral/framework/pull/733

kafkiansky:
@butschster, added tests.

wolfy-j:
Looks cool!

roxblnfk (commented):
I think the rendering is not a responsibility of a Filter.
Idea about rendering customization is great but it should be not in a filter.

kafkiansky:
> I think the rendering is responsibility of a Filter.
> Idea about rendering customization is great but it should be not in a filter.

I think whoever is owned the data and the validation rules for that data, and may know how to output the error information. Otherwise we would have to keep this behaviour separate, which would complicate the creation and maintenance of filters. Moreover, this behaviour is optional.

wolfy-j:
Maybe just let interceptor to know which renderer to use via attribute on filter?

kafkiansky:
> Maybe just let interceptor to know which renderer to use via attribute on filter?

I don't think such a small feature should require that much boilerplate. In that case you need:

1. A specific interface that be able to render a validation errors:

```php
<?php

declare(strict_types=1);

use Spiral\Filters\Filter;

/**
 * @psalm-template F of Filter
 */
interface RenderErrors
{
     /**
      * @psalm-param F $filter
      *
      * @return mixed
      */
     public function render(Filter $filter);
}
```

2. An attribute:

```php
<?php

declare(strict_types=1);

#[\Attribute(\Attribute::TARGET_CLASS)]
final class RenderWith
{
    /**
     * @param class-string<RenderErrors> $rendererFqcn
     */
    public function __construct(
          public readonly string $rendererFqcn,
    ) {}
}
```

3. Some locator that be able to collect the renderable filters to avoid using reflection in runtime each time.

The only advantage of this approach that I have found is that such classes will be autowired:

```php
<?php

declare(strict_types=1);

use Spiral\Filters\Filter;
use Psr\Log\LoggerInterface;

/**
 * @template-implements RenderErrors<Filter>
 */
final class SomeSpecificRenderer implements RenderErrors
{
     public function __construct(
          private readonly LoggerInterface $logger,
     ) {}

     /**
      * {@inheritdoc}
      */
     public function render(Filter $filter)
     {
           $this->logger->info(...);
           
           return [...];
     }
}
```

I don't think this approach is much cleaner than the one I suggested. But definitely not worse. 

I understood your idea correctly?

kafkiansky:
I updated the PR, but now it has added BC.

roxblnfk (approved):

roxblnfk:
Separated interceptor and annotation look better

roxblnfk (approved):

Issue #572, opened 2022-02-17
Classes locator optimization


butschster:
We have one performance  problem with using ClasssesInterface. A lot of components or packages use this class for specific class location. And every package scan files again and again.

It would be great if SpiralFramework has an ability to scan all required directories and to notify all subscribers about found classes.

kafkiansky:
Hi, @butschster. I'd like to help with this feature, because we use a lot of tokenization, which slows down bootstrap applications and tests. I have ideas about this, which I have implemented in my project, but I will not risk doing the PR, as you may have a completely different implementation. But I'll describe it in this issue.

1. First, we need an interface for tokenization listeners. It might look something like this:

```php
<?php

declare(strict_types=1);

interface TokenizationListenerInterface
{
    public function listen(\ReflectionClass $r): void;
    public function finalize(): void;
}
```

The `finalize` method is needed, first, so that listeners can clear their state if they have been accumulating it, and second, so that they can pass the accumulated objects to some service if it is waiting for them in the constructor, for example.

2. After that, we need some locator with which we can collect the listeners interested in tokenization and invoke them at the moment of tokenization.

```php
<?php

declare(strict_types=1);

final class TokenizationListenerLocator
{
    /**
     * @var TokenizationListenerInterface[]
     */
    private array $tokenizationListeners = [];

    public function addListener(TokenizationListenerInterface $tokenizationListener): void
    {
        $this->tokenizationListeners[] = $tokenizationListener;
    }

    public function invokeListeners(\ReflectionClass $r): void
    {
        foreach ($this->tokenizationListeners as $tokenizationListener) {
            $tokenizationListener->listen($r);
        }
    }
    
    public function finalize(): void
    {
        foreach ($this->tokenizationListeners as $tokenizationListener) {
            $tokenizationListener->finalize();
        }

        unset($this->tokenizationListeners);
    }
}
```

3. Third, we need a point where we do the tokenization and notify the listeners. To do this we can use `AbstractKernel` callbacks, but for that we need a different type of callback, since we need to wait for the bootstrap of the application. Let's call them, for example, `bootrstrappedCallbacks` and call them right before the application `serve` method. 

```php
<?php

declare(strict_types=1);

abstract class AbstractKernel
{
    ...
    
    /** @var array<Closure>  */
    private $bootrstrappedCallbacks = [];
   
    public function __construct()
    {
         ...
         $this->container->bindSingleton(TokenizationListenerLocator::class, new TokenizationListenerLocator());
    }
[...]

PR #763, opened 2022-08-25
Tokenizer listeners

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️
| Issues        | #572

- [x] Tokenizer listener
- [x] Unit tests

With this PR you will have the ability to use `Spiral\Tokenizer\Bootloader\TokenizerListenerBootloader` instead of `Spiral\Tokenizer\Bootloader\TokenizerBootloader`. It will iterate only once all classes after application bootstrapping all bootloaders from `LOAD` and `APP` sections and at first send every found class into all registered via bootloaders listeners and after that will finalize the listeners where you can handle all matched classes.

It will boost application bootstrapping performance.

Example of usage
```php
<?php

declare(strict_types=1);

namespace App\Bootloader;

use Spiral\Boot\Bootloader\Bootloader;
use Spiral\Bootloader\Attributes\AttributesBootloader;
use Spiral\Tokenizer\Bootloader\TokenizerListenerBootloader;

/**
 * Configures application routes using annotations and pre-defined configuration groups.
 */
final class MyBootloader extends Bootloader
{
    protected const DEPENDENCIES = [
        AttributesBootloader::class,
        TokenizerListenerBootloader::class,
    ];

    public function init(
        TokenizerListenerBootloader $tokenizer,
        MyClassesLocatorListener $listener
    ): void {
        $tokenizer->addListener($listener);
    }
}
```

```php
<?php

declare(strict_types=1);

namespace Spiral\App;

use ReflectionClass;
use Spiral\Attributes\ReaderInterface;
use Spiral\Tokenizer\TokenizationListenerInterface;

final class MyClassesLocatorListener implements TokenizationListenerInterface
{
    public function __construct(
        private readonly ReaderInterface $reader,
        private readonly MyClassesRegistry $registry
    ) {
    }

    public function listen(ReflectionClass $class): void
    {
        $attr = $this->reader->firstFunctionMetadata($method, MyAttribute::class);

        if ($attr === null) {
            return;
        }

        $this->registry->register($attr ->name, $class->getName());
    }

    public function finalize(): void
    {
        $this->registry->...
    }
}
```


src/Prototype/src/Traits/PrototypeTrait.php
```diff
@@ -11,6 +11,41 @@
 
 /**
  * This DocComment is auto-generated, do not edit or commit this file to repository.
+ *
```
butschster, line 14: Should be removed

msmakouz (approved):

roxblnfk (approved):

PR #826, opened 2022-10-26
Create TokenStorageProvider to able manage token storages

kastahov:
| Q             | A
| ------------- | ---
| New feature?  | ✔️

TokenStorageProvider provides a convinient way to manage token storages for authentication

**Token storage registration example**
```php
use Spiral\Boot\Bootloader;
use Spiral\Bootloader\Auth\HttpAuthBootloader;

final class SomeBootloader extends Bootloader
{
    public function init(HttpAuthBootloader $httpAuth, DatabaseTokenStorage $storage): void 
    {
        $httpAuth->addTokenStorage('database', $storage);
    }
}
```

**`config/auth.php` example**

```PHP
use Spiral\Core\Container\Autowire;

return [
    'defaultTransport' => env('AUTH_TOKEN_TRANSPORT', 'cookie'),
    'defaultStorage' => env('AUTH_TOKEN_STORAGE', 'session'),
    'transports' => [],
    'storages' => [
         'session' => \Spiral\Auth\Session\TokenStorage::class,
         'database' => new Autowire( \App\DataaseTokenStorage::class, ['table' => 'auth_tokens']),
         'jwt' => new Autowire( \App\JwtTokenStorage::class, ['key' => 'secret']),
         'cycle-orm' => 'cycleorm.storage' // Will be requested from the container
    ]
]
```

You can use`AUTH_TOKEN_STORAGE`  variable in `.env`. There you can set a default stoage.
```PHP
AUTH_TOKEN_STORAGE=session
```

**RoutersBootloader example**
```PHP
use Spiral\Core\Container\Autowire;
use Spiral\Auth\Middleware\AuthTransportWithStorageMiddleware;

return [
    'web' => [
        new Autowire(
            AuthTransportWithStorageMiddleware::class,
            ['transportName' => 'cookie', 'storage' => 'session']
        ),
    ],
    'api' => [
        new Autowire(
            AuthTransportWithStorageMiddleware::class,
            ['transportName' => 'header', 'storage' => 'jwt']
        ),
    ],
];
```


butschster (changes requested):

butschster (approved):

msmakouz:
I'm not sure if we need a separate `AuthTransportWithStorageMiddleware`.

It might be worth updating `AuthTransportMiddleware`:
1) Extend the type `TokenStorageInterface $tokenStorage` to `TokenStorageInterface|string $tokenStorage`
2) Add an optional parameter `TokenStorageProviderInterface` to the constructor.
3) In the `AuthTransportMiddleware` code, check if a string is passed and `TokenStorageProviderInterface` is not null (if it is null and storage string - throw an exception) - get the required storage from storage by its string name.

In version 4.0, we could make TokenStorageProviderInterface a required parameter and we can remove checking and exception.

msmakouz (approved):

PR #869, opened 2023-01-30
Adding the ability to configure a prefix in the storage alias

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

Added the ability to configure prefixes in the storage alias.

### app/config/cache.php

```php
use Spiral\Cache\Storage\ArrayStorage;

return [
    'aliases' => [
        'user-data' => 'localMemory',
        'blog-data' => [
            'storage' => 'localMemory',
            'prefix' => 'blog_'
        ],
    ],
    'storages' => [
        'localMemory' => [
            'type' => ArrayStorage::class,
        ],
    ],
];
```

### Setting data

```php
// CacheManager $cache
$cache->storage('user-data')->set('data', 'foo');
$cache->storage('blog-data')->set('data', 'bar');
```

### Result

```php
array:2 [
  "data" => array:2 [
    "value" => "foo"
    "timestamp" => 1677694480
  ]
  "blog_data" => array:2 [
    "value" => "bar"
    "timestamp" => 1677694480
  ]
]
```

### Bugfix

Added missed **EventDispatcher** when creating **CacheRepository** and the test for this case.

### DIFF
--- a/src/Cache/src/CacheManager.php
+++ b/src/Cache/src/CacheManager.php
@@ -35,11 +35,11 @@ class CacheManager implements CacheStorageProviderInterface, SingletonInterface
             $storage = $storage['storage'];
         }
 
-        if (isset($this->storages[$storage])) {
-            return new CacheRepository($this->storages[$storage], $this->dispatcher, $prefix);
+        if(!isset($this->storages[$storage])) {
+            $this->storages[$storage] = $this->resolve($storage);
         }
 
-        return new CacheRepository($this->storages[$storage] = $this->resolve($storage), $this->dispatcher, $prefix);
+        return new CacheRepository($this->storages[$storage], $this->dispatcher, $prefix);
     }
 
     private function resolve(?string $name): CacheInterface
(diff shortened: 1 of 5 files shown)


src/Cache/src/CacheManager.php
```diff
@@ -25,19 +27,25 @@ public function storage(?string $name = null): CacheInterface
...
-        if (isset($this->storages[$name])) {
-            return $this->storages[$name];
+        $prefix = null;
+        if (\is_array($storage)) {
+            $prefix = !empty($storage['prefix']) ? $storage['prefix'] : null;
+            $storage = $storage['storage'];
         }
 
-        return $this->storages[$name] = $this->resolve($name);
+        if (isset($this->storages[$storage])) {
```
butschster, line 38:
```php
if(!isset($this->storages[$storage])) {
    $this->storages[$storage] = $this->resolve($storage);
}

return new CacheRepository($this->storages[$storage], $this->dispatcher, $prefix);
```

butschster (changes requested):

src/Cache/src/CacheManager.php
msmakouz: Fixed, thanks

butschster (approved):

PR #871, opened 2023-02-07
Ignore exception handling for already silent errors

roquie:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | NO 

## Problem

I want to temporary hide deprecated errors and I go to `app.php` to set custom error reporting:
```php
\error_reporting(E_ERROR | E_WARNING | E_PARSE | E_NOTICE);
// or \error_reporting(E_ALL ^ E_DEPRECATED);
```

and then I ran my cli script again. I still got an error with deprication but wrapped by framework error handler.

## Solution

Don't handle PHP errors with the framework's error handler unless the developer wants to.

### DIFF
--- a/src/Exceptions/src/ExceptionHandler.php
+++ b/src/Exceptions/src/ExceptionHandler.php
@@ -129,17 +129,23 @@ class ExceptionHandler implements ExceptionHandlerInterface
      */
     protected function handleShutdown(): void
     {
-        if (!empty($error = \error_get_last())) {
-            $this->handleGlobalException(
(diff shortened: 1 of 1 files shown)


roxblnfk (approved):
That's OK but can you add few tests about it?
The tests should be run in a separated processes

roquie:
@roxblnfk 👍

butschster:
@roquie Thank you!

PR #895, opened 2023-03-14
Fixed the problem with using named parameters in class located by a tokenizer.

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌

The Tokenizer component had a bug that caused issues when trying to load a class and parse tokens. Specifically, the bug occurred when a class contained a `class` named parameter, as demonstrated in the example code below.

```php
namespace Spiral\Tests\Tokenizer\Classes;

use Spiral\Tests\Tokenizer\Classes\Inner\ClassD;

class ClassWithNamedParameter
{
    public function __construct()
    {
        $this->foo(class: ClassD::class);
    }

    private function foo(string $class): void
    {
    }
}
```

When the Tokenizer encountered this scenario, it incorrectly counted `ClassD` as part of `ClassWithNamedParameter` in the same namespace `Spiral\Tests\Tokenizer\Classes`.

Thanks to @gam6itko

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## Unreleased
+- **Medium Impact Changes**
+  - [spiral/queue] Added the ability to use mixed types as job payload.
+- **Bug Fixes**
+  - [spiral/console] Fixed the problem with commands description with signature definition.
+  - [spiral/tokenizer] Fixed the problem with using named parameters in class located by a tokenizer.
+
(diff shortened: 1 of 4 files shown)


gam6itko:
@butschster can you also add tests for anonymous classes ?

```php
new class ($arg1, $arg2) implements SomeInterface {
            
    private function someFunc(): void
    {
    }

}
```

msmakouz (approved):

roxblnfk (approved):

PR #902, opened 2023-03-22
Scaffolder improvements

butschster:
### Improves create:bootloader command
- Adds final for class
- Adds an ability to create domain bootloader with interceptors using `-d` option
- Uses PHP attributes for command definition

```bash
// Generate domain bootloader
php app.php create:bootloader AppBootloader -d
```

### DIFF
--- a/src/Scaffolder/src/Command/AbstractCommand.php
+++ b/src/Scaffolder/src/Command/AbstractCommand.php
@@ -31,7 +31,7 @@ abstract class AbstractCommand extends Command
      * @param class-string<TClass> $class
      * @return TClass
      */
-    protected function createDeclaration(string $class): DeclarationInterface
+    protected function createDeclaration(string $class, array $params = []): DeclarationInterface
     {
         return $this->factory->make(
             $class,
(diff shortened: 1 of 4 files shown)


src/Scaffolder/src/Declaration/CommandDeclaration.php
```diff
@@ -26,18 +57,34 @@ public function setDescription(string $description): void
...
+            ->addMethod('__invoke')
+            ->setReturnType('int')
+            ->setBody(
+                <<<'PHP'
+// Put your command logic here
+$this->info('Command logic is not implemented yet');
+
+return self::SUCCESS;
+PHP,
+            );
```
roxblnfk, line 88:
```suggestion
                <<<'PHP'
                // Put your command logic here
                $this->info('Command logic is not implemented yet');

                return self::SUCCESS;
                PHP,
            );
```

src/Scaffolder/src/Declaration/ConfigDeclaration.php
```diff
@@ -65,13 +65,19 @@ public function declare(): void
...
             ->setProtected()
             ->setType('array')
             ->setValue([])
-            ->setComment('@internal For internal usage. Will be hydrated in the constructor.');
+            ->setComment(
+                <<<'DOC'
+Default values for the config. 
+Will be merged with application config in runtime.
+DOC,
+            );
```
roxblnfk, line 80:
```suggestion
            ->setComment(
                <<<'DOC'
                Default values for the config. 
                Will be merged with application config in runtime.
                DOC,
            );
```

roxblnfk (changes requested):

PR #907, opened 2023-03-28
Fixes queue payload type from array to mixin

butschster:
- Fixes unit tests
- Marks some classes as deprecated

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌


src/Queue/src/Job/CallableJob.php
```diff
@@ -8,6 +8,9 @@
 use Spiral\Queue\Exception\InvalidArgumentException;
 use Spiral\Queue\HandlerInterface;
 
+/**
+ * @deprecated since v4.0
```
msmakouz, line 12: Since v4.0?

PR #955, opened 2023-07-18
[spiral/core] Throw an exception on rebinding of a resolved (cached) singleton

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 

### What was changed

1. Fixed `removeBinding` method in the `Spiral\Core\Internal\Config\StateBinder`, added removal of alias from `singletons` property.
2. Added an exception when trying to override an existing singleton. If need to redefine, a user must first explicitly remove the current singleton definition.

### DIFF
--- a/src/Boot/src/Bootloader/ConfigurationBootloader.php
+++ b/src/Boot/src/Bootloader/ConfigurationBootloader.php
@@ -49,6 +49,8 @@ final class ConfigurationBootloader extends Bootloader
     {
         if (!isset($this->loaders[$ext]) || $this->loaders[$ext]::class !== $loader::class) {
             $this->loaders[$ext] = $loader;
+
+            $this->binder->removeBinding(ConfigManager::class);
             $this->binder->bindSingleton(ConfigManager::class, $this->createConfigManager());
         }
     }
--- a/src/Boot/src/Bootloader/ConfigurationBootloader.php
+++ b/src/Boot/src/Bootloader/ConfigurationBootloader.php
@@ -49,8 +49,6 @@ final class ConfigurationBootloader extends Bootloader
     {
         if (!isset($this->loaders[$ext]) || $this->loaders[$ext]::class !== $loader::class) {
             $this->loaders[$ext] = $loader;
-
-            $this->binder->removeBinding(ConfigManager::class);
             $this->binder->bindSingleton(ConfigManager::class, $this->createConfigManager());
         }
     }
--- a/src/Core/src/Exception/Binder/DuplicateBindingException.php
+++ b/src/Core/src/Exception/Binder/DuplicateBindingException.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Spiral\Core\Exception\Binder;
+
+use Spiral\Core\Exception\ConfiguratorException;
+
+final class DuplicateBindingException extends ConfiguratorException
+{
+}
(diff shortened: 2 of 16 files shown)


butschster (approved):

src/Core/src/Exception/Binder/DuplicateBindingException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Spiral\Core\Exception\Binder;
+
+use Spiral\Core\Exception\ConfiguratorException;
+
+final class DuplicateBindingException extends ConfiguratorException
```
roxblnfk, line 9:
```suggestion
final class SingletonOverloadException extends ConfiguratorException
```
Smth like this.

roxblnfk, line 9: Please describe the solution with `removeBinding` in the phpdoc

src/Core/src/Internal/Config/StateBinder.php
```diff
@@ -58,6 +59,13 @@ public function bind(string $alias, mixed $resolver): void
      */
     public function bindSingleton(string $alias, mixed $resolver): void
     {
+        if ($this->hasInstance($alias)) {
+            throw new DuplicateBindingException(\sprintf(
+                'Singleton `%s` already exists. To override a binding, call the `removeBinding` method first.',
```
roxblnfk, line 64: The message should be in the exception class and shouldnt contain a solution.
"Can't overload the singleton `%s` because it's already used."

src/Boot/src/Bootloader/ConfigurationBootloader.php
```diff
@@ -49,6 +49,8 @@ public function addLoader(string $ext, FileLoaderInterface $loader): void
     {
         if (!isset($this->loaders[$ext]) || $this->loaders[$ext]::class !== $loader::class) {
             $this->loaders[$ext] = $loader;
+
+            $this->binder->removeBinding(ConfigManager::class);
```
roxblnfk, line 53: @butschster is that rebinding OK?

roxblnfk:
We can also add `$force` parameter to `bindSingleton` and make it `true` by default with note that it will be `false` in the future

src/Core/src/Internal/Config/StateBinder.php
```diff
@@ -58,6 +59,13 @@ public function bind(string $alias, mixed $resolver): void
      */
     public function bindSingleton(string $alias, mixed $resolver): void
     {
+        if ($this->hasInstance($alias)) {
+            throw new DuplicateBindingException(\sprintf(
```
roxblnfk, line 63: You must check the existing in cache. Does the method `hasInstance` do the same?

src/Core/src/Exception/Binder/DuplicateBindingException.php
msmakouz: Thanks, renamed.

src/Core/src/Internal/Config/StateBinder.php
msmakouz: Thanks, changed.

src/Core/src/Exception/Binder/DuplicateBindingException.php
msmakouz: Added information about removeBinding

roxblnfk (approved):

PR #963, opened 2023-08-10
Changes behavior for loading system bootloaders. 

butschster:
Now system bootloaders init only after Environment is bound to the container

---

This PR also adds Tokenizer info command

image


msmakouz (approved):

msmakouz:
@butschster Maybe, now we can stop using **running** here: https://github.com/spiral/framework/blob/master/src/Bridge/Dotenv/src/Bootloader/DotenvBootloader.php#L21

Issue #988, opened 2023-09-07
[spiral/router] Issue with Registering Routes Containing Host using RoutesBootloader


butschster:
When using the `RoutesBootloader` to define routes that begin with a host (using `//<host>`), the `RoutingConfigurator` seems to trim the starting double slashes. This leads to the `Router` misunderstanding and registering the pattern as if it started from the path, not the host.

**Steps to reproduce:**

1. Define a route using `RoutesBootloader` in the following manner:

```php
protected function defineRoutes(RoutingConfigurator $routes): void
{
    $routes->add('register', "//<host>/register")
           ->callable(fn (string $host) => $host);
}
````

2. Check how the route is registered in the Router.

**Expected behavior:**
The route should be registered as `//<host>/register`, indicating that the pattern starts with the host name.

**Actual behavior:**
The Router trims the starting double slashes and registers the route as `<host>/register`, causing it to misinterpret the route pattern.

**Additional notes:**
Using PHP attributes for route registration does not seem to exhibit this issue; it's specific to RoutesBootloader.

PR #990, opened 2023-09-11
[spiral/router] Adding host check when importing routes via RoutingConfigurator

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | #988 

### What's was changed

Fixed an error when importing routes with a pattern starting with **host**. For example:
```php
protected function defineRoutes(RoutingConfigurator $routes): void
{
    $routes->add('test', '//<host>/register-configurator')->callable(fn () => 'foo');
}
```

Before changes:

After changes:

### DIFF
--- a/src/Router/src/Router.php
+++ b/src/Router/src/Router.php
@@ -147,7 +147,10 @@ final class Router implements RouterInterface
                 $target = $target->withCore($configurator->core);
             }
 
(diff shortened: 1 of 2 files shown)


butschster (approved):

rustatian (approved):
LGTM

PR #1036, opened 2023-12-20
[spiral/console] Add the ability to use enum as console option

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

The ability to use Enum as an option in a console command when configuring it with attributes has been added. For example, option Status in the example below:

```php
use App\Domain\User\Status;
use Spiral\Console\Attribute\Argument;
use Spiral\Console\Attribute\AsCommand;
use Spiral\Console\Attribute\Option;
use Spiral\Console\Attribute\Question;
use Spiral\Console\Command;

#[AsCommand(name: 'app:create-user', description: 'Create user')]
final class CreateUser extends Command
{
    #[Argument(description: 'User email')]
    #[Question(question: 'Provide user email')]
    private string $email;

    #[Option(description: 'User status')]
    private Status $status;

    public function __invoke(): int
    {
        // ...
    }
}
```

### DIFF
--- a/src/Console/src/Configurator/Attribute/Parser.php
+++ b/src/Console/src/Configurator/Attribute/Parser.php
@@ -179,12 +179,15 @@ final class Parser
             return $value;
         }
 
-        return match ($type->getName()) {
-            'int' => (int) $value,
-            'string' => (string) $value,
-            'bool' => (bool) $value,
-            'float' => (float) $value,
-            'array' => (array) $value,
+        $typeName = $type->getName();
+
+        return match (true) {
+            $typeName === 'int' => (int) $value,
+            $typeName === 'string' => (string) $value,
+            $typeName === 'bool' => (bool) $value,
+            $typeName === 'float' => (float) $value,
+            $typeName === 'array' => (array) $value,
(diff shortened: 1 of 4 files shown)


src/Console/src/Configurator/Attribute/Parser.php
```diff
@@ -179,12 +179,15 @@ private function typecast(mixed $value, \ReflectionProperty $property): mixed
...
-            'array' => (array) $value,
+        $typeName = $type->getName();
+
+        return match (true) {
+            $typeName === 'int' => (int) $value,
+            $typeName === 'string' => (string) $value,
+            $typeName === 'bool' => (bool) $value,
+            $typeName === 'float' => (float) $value,
+            $typeName === 'array' => (array) $value,
+            \enum_exists($typeName) => $typeName::from($value),
```
butschster, line 190: If the console option receives an incorrect value, not allowed by the PHP enum, we should display an error to the user like this: 'Wrong option value. Allowed options: 'foo', 'bar'.

butschster (changes requested):

PR #1041, opened 2023-12-22
Extended injector

### LINKED ISSUE #888: [spiral/core] Injectors context
We can pass a context into an injector resolver, where the context is a string, usually method argument name. I would be great if the context was a Parameter object and we could get attributes for it or get information about class, etc. In this case we will have more flexible injectors

roxblnfk:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌
| New feature?  | ✔️
| Issues        | Closes #888

## Container Factory

Now the container factory accepts a `Stringable` context. This can be either reflection or any other `Stringable`.

```php
$container->make('foo', new StringableContext());
```

Other container methods with an extended context will be available in a separate PR.

## Injectors

If the injector has an extended signature that allows passing an extended context (type of the `$context` parameter is `Stringable|string|null`, `mixed` or `ReflectionParameter|string|null`), the extended context will be passed.

```php
class ExtendedContextInjector implements InjectorInterface
{
    public function createInjection(\ReflectionClass $class, \ReflectionParameter|string|null $context = null): object
    {
        return (object)['context' => $context];
    }
}
```

### DIFF
--- a/src/Core/src/Attribute/Singleton.php
+++ b/src/Core/src/Attribute/Singleton.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Spiral\Core\Attribute;
 
-use Spiral\Core\Internal\Factory\Ctx;
-
 /**
  * Mark class as singleton.
  */
--- a/src/Core/src/Container.php
+++ b/src/Core/src/Container.php
@@ -114,7 +114,7 @@ final class Container implements
      * @throws ContainerException
      * @throws \Throwable
      */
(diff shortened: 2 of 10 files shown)


roxblnfk:
> 2627 Segmentation fault      (core dumped) vendor/bin/phpunit --coverage-clover=coverage.clover

😟

wolfy-j (approved):

butschster (approved):

Issue #650, opened 2022-04-02
Autowiring with proxy


roxblnfk:
## Hard way

Using type association and intersection with the Lazy class, we can generate proxies and decorators.

```php
class Service {
    public function __construct(
        Lazy|OrmInterface $proxy, // magic proxy
        Lazy&OrmInterface $decorator,
    ) {
        \assert($proxy instanceof Lazy);
        \assert($proxy::class === Lazy::class || $proxy::class === Proxy::class);
        \assert(!$proxy instanceof OrmInterface);

        \assert($decorator::class === 'AutoGeneratedDecoratorClassName');
        \assert($decorator instanceof Lazy);
        \assert($decorator instanceof OrmInterface);

        $proxy->getSchema();
        $decorator->getSchema();

        # Problems
        ## There are the same problems with properties like in Cycle ORM proxy entities
            $proxy->list[] = 'foo'; // error if proxy is not resolved
            $foo = &$proxy->list;   // error
        ## Decorator will have very difficult code
        ## Decorator is non-compatible with final classes and methods
    }
}
```

## Simple way

Make the proxy without magic, just with `get()` method.

```php
/**
 * @template TDefinition
 */
class Proxy {
    /**
     * @return TDefinition
     */
    public function get(): mixed;
}
class Service {
    /**
     * @var Proxy<OrmInterface> Use template for autocompletion
     */
    private Proxy $orm;

    public function __construct(
        Proxy|OrmInterface $proxy,
    ) {
        $this->orm = $proxy; // Mb psalm will throw some warning here
    }

    public function doSomeAction() {
        $this->orm->get() // Resolve ORM
            ->getSchema();
        // ...
    }
}
```

wolfy-j:
Love it. We can also use attributes.

PR #1045, opened 2023-12-28
Scope proxy

roxblnfk:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️ 
| Issues        | Similar to #650
| Docs PR       | 

# What was changed

Note: Includes changes from https://github.com/spiral/framework/pull/1041

## Context

The context is also extended on other container methods `get()` (see https://github.com/spiral/framework/pull/1041)
### Default scope fix

If the container scope is not open, it is assumed by default that dependencies are resolved in the scope named `root`. Now when calling `invoke()`, `make()`, `get()`, the container will globally register itself with the `root` scope if no other scope was opened. Before this, the container resolved dependencies as if outside the scope.

### Scoped Interface

The experimental `ContainerScopeInterface` has been removed. The method `getBinder(?string $scope = null): BinderInterface` has been moved to `BinderInterface` at the annotation level.

### runScope method

The `Container::runScoped()` method (in the implementation) was additionally marked as `@deprecated` and will be removed when its use in tests is reduced to zero. Instead of the `Container::runScoped()`, you should now call the old `Container::runScope()`, but with passing the DTO `Spiral\Core\Scope` instead of the list of bindings.

```php
$container->runScope(
    new Scope(name: 'auth', bindings: ['actor' => new Actor()]),
    function(ContainerInterface $container) {
        dump($container->get('actor'));
    },
);
```

### Scope Proxy

Instead of the now removed `ContainerScopeInterface::getCurrentContainer()` method, the user is offered another way to get dependencies from the container of the current scope - a proxy.

The user can mark the dependency with a new attribute `Spiral\Core\Attribute\Proxy`.
Important: the dependency must be defined by an interface.
When resolving dependencies, the container will create a proxy object that implements the specified interface. When calling the interface method, the proxy object will get the container of the current scope, request the dependency from it using its interface, and start the necessary method.

```php
final class Service  
{
    public function __construct(  
        #[Proxy] public LoggerInterface $logger,  
    ) {  
    }

    public function doAction() {
        // Equals to
        // $container->getCurrentContainer()->get(LoggerInterface::class)->log('foo')
        $this->logger->log('foo'); 
    }
}
```

Important nuances:
- The proxy refers to the active scope of the container, regardless of the scope in which the proxy object was created.
- Each call to the proxy method pulls the container. If there are many calls within the method, you should consider making a proxy for the container
   ```php
   // class
   function __construct(
       #[Proxy] private Dependency $dep,
       #[Proxy] private ContainerInterface $container,
   ) {}
   function handle() {
       // There are four calls to the container under the hood.
       $this->dep->foo();
       $this->dep->bar();
       $this->dep->baz();
       $this->dep->red();
       
       // Only two calls to the container and caching the value in a variable
       // The first call - getting the container through the proxy
       // The second - explicit retrieval of the dependency from the container
       $dep = $this->container->get(Dependency::class);
       $dep->foo();
       $dep->bar();
       $dep->baz();
       $dep->red();
   }
   ```
- The proxied interface should not contain a constructor signature (although this sometimes happens).
- Calls to methods outside the interface will not be proxied. This option is possible in principle, but it is disabled. If it is absolutely necessary, we will consider whether to enable it.
- The destructor method call will not be proxied.


butschster (approved):

butschster:
@roxblnfk Couc you please add some example of usage proxy attribute?

roxblnfk:
> @roxblnfk Could you please provide some example of usage proxy attribute?

done

PR #1061, opened 2024-01-15
[spiral/core] Add the ability to specify the scope of the dispatcher using an attribute

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

- [x] https://github.com/spiral/testing/pull/70

### DispatcherScope

Added the ability to specify the scope name for the dispatcher using the `Spiral\Attribute\DispatcherScope` attribute.

```php
use Spiral\Attribute\DispatcherScope;
use Spiral\Boot\DispatcherInterface;
use Spiral\Framework\ScopeName;

#[DispatcherScope(scope: ScopeName::Console)]
final class ConsoleDispatcher implements DispatcherInterface
{
    // ...
}
```

### Registration of dispatchers

The registration of dispatchers has been changed. The accepted type in the **addDispatcher** method of the `Spiral\Boot\AbstractKernel` class has been extended from **DispatcherInterface** to **string|DispatcherInterface**. Before these changes, the method accepted a created DispatcherInterface object, now it can accept a class name string or an object. In version 4.0, the **DispatcherInterface** type will be removed. When passing an object, only its class name will be saved. And when using the dispatcher, its object will be created anew. Example with ConsoleDispatcher:

```php
public function init(AbstractKernel $kernel): void
{
    $kernel->bootstrapped(static function (AbstractKernel $kernel): void {
        $kernel->addDispatcher(ConsoleDispatcher::class);
    });
}
```

### Using dispatchers

The dispatchers are now created in their own scope and receive dependencies that are specified in this scope. But due to the need to check whether the dispatcher can handle the request or not before creating the dispatcher object, the **canServe** method in dispatchers must be **static**:

```php
public static function canServe(EnvironmentInterface $env): bool
{
    return (PHP_SAPI === 'cli' && $env->get('RR_MODE') === null);
}
```

This method has been removed from the `Spiral\Boot\DispatcherInterface`, for backward compatibility it can be non-static, as it was before (then an object will be created for its call) or static and accept `Spiral\Boot\EnvironmentInterface`.


wolfy-j:
All good, but please hit me to check naming before release

butschster (changes requested):

src/Boot/src/AbstractKernel.php
```diff
@@ -278,21 +287,31 @@ public function serve(): mixed
...
-                        $eventDispatcher?->dispatch(new DispatcherFound($dispatcher));
-                        return $dispatcher->serve();
-                    }
-                );
+            $reflection = new \ReflectionClass($dispatcher);
+
+            $scope = ($reflection->getAttributes(DispatcherScope::class)[0] ?? null)?->newInstance()->scope;
+            $this->container->getBinder($scope)->bind($dispatcher, $dispatcher);
+
+            if ($serving === null && $this->canServe($reflection)) {
```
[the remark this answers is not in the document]
msmakouz: @butschster We can't do this because we bind all dispatchers in the container from the $this->dispatchers property. Therefore, the loop cannot be interrupted.

src/Boot/src/AbstractKernel.php
```diff
@@ -259,8 +264,12 @@ public function bootstrapped(Closure ...$callbacks): void
      * Add new dispatcher. This method must only be called before method `serve`
      * will be invoked.
      */
-    public function addDispatcher(DispatcherInterface $dispatcher): self
+    public function addDispatcher(string|DispatcherInterface $dispatcher): self
```
roxblnfk, line 267: Please describe the `$dispatcher` parameter

src/Boot/src/DispatcherInterface.php
```diff
@@ -7,14 +7,10 @@
 /**
  * Dispatchers are general application flow controllers, system should start them and pass exception
  * or instance of snapshot into them when error happens.
+ * @method static bool canServe()
```
roxblnfk, line 10:
```suggestion
 * @method static bool canServe() Must return true if the dispatcher expects to handle requests in a current environment
```

roxblnfk (changes requested):

src/Boot/src/AbstractKernel.php
msmakouz: Added

src/Boot/src/DispatcherInterface.php
msmakouz: Added

butschster (approved):

PR #1069, opened 2024-01-23
[spiral/router] Opening of `http.request` scope added

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

## What was changed

Opening of `http.request` scope added, `Proxy` attributes added to get dependencies in the correct scope.


src/AnnotatedRoutes/tests/IntegrationTest.php
```diff
@@ -85,8 +91,10 @@ public function post(
         array $headers = [],
         array $cookies = []
     ): ResponseInterface {
-        return $this->app->getHttp()->handle(
-            $this->request($uri, 'POST', [], $headers, $cookies)->withParsedBody($data)
+        return $this->app->getContainer()->runScope(
+            new Scope(ScopeName::Http),
```
wolfy-j, line 95: we will come up with better name for the enum before release

src/AnnotatedRoutes/tests/IntegrationTest.php
```diff
@@ -63,7 +66,10 @@ public function get(
         array $headers = [],
         array $cookies = []
     ): ResponseInterface {
-        return $this->app->getHttp()->handle($this->request($uri, 'GET', $query, $headers, $cookies));
+        return $this->app->getContainer()->runScope(
+            new Scope(ScopeName::Http),
+            fn (Http $http) => $http->handle($this->request($uri, 'GET', $query, $headers, $cookies))
```
wolfy-j, line 71: i wonder if we can have container in HTTP scope already here... @roxblnfk like $this->getHttpScopeContainer() and then we work from there. can we do that?

src/Hmvc/tests/CoreTest.php
```diff
@@ -7,191 +7,232 @@
...
     {
-        $core = new SampleCore(new Container());
-        $this->assertSame('Hello, Antony.', $core->callAction(
-            DummyController::class,
-            'index',
-            ['name' => 'Antony']
-        ));
+        $core = new SampleCore($this->root);
+
+        $this->root->runScope(new Scope(ScopeName::Http), function () use ($core): void {
```
wolfy-j, line 29: we have to simplify this for users, it's ok in framework, but people won't be creating HTTP scope by themselves.

wolfy-j: also, the test can be in TEST scope overall, most likely "HTTP" scope won't be doing much, HTTP.request will.

src/Hmvc/tests/CoreTest.php
```diff
@@ -7,191 +7,232 @@
...
     public function testCallNotController(): void
     {
         $this->expectException(ControllerException::class);
 
-        $core = new SampleCore(new Container());
-        $core->callAction(SampleCore::class, 'index', [
-            'name' => 'Antony',
-        ]);
+        $core = new SampleCore($this->root);
+        $this->root->runScope(new Scope(ScopeName::Http), function () use ($core): void {
```
wolfy-j, line 150: yeah, too much overhead

src/AnnotatedRoutes/tests/IntegrationTest.php
roxblnfk: I think about using attributes for scoped tests:

```php
#[TestScope('http.request')]
public function testSomethingInHttpRequest() {...}
```

tests/Framework/Bootloader/Http/HttpBootloaderTest.php
```diff
@@ -21,7 +24,17 @@ final class HttpBootloaderTest extends BaseTestCase
 {
     public function testHttpBinding(): void
     {
-        $this->assertContainerBoundAsSingleton(Http::class, Http::class);
+        $this->getContainer()->runScope(new Scope(ScopeName::Http), function (Container $container): void {
```
roxblnfk, line 27: Looks like you can use TestScope attribute here

msmakouz: Yes, you're right

src/AnnotatedRoutes/tests/IntegrationTest.php
msmakouz: @wolfy-j @roxblnfk Added **TestScope** attribute to the spiral/testing package, simplified tests.

roxblnfk (approved):

PR #1073, opened 2024-02-03
Update TranslatorConfig.php

lee-to:
After installation or after executing the 'php app.php configure' command, the directory with the locale is not created correctly
Снимок экрана 2024-02-03 в 19 11 28

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌


msmakouz:
@lee-to Hi, thank you for the work you have done. Fixed by #1075

PR #1077, opened 2024-02-15
[spiral/core] Added the ability to bind the interface as a proxy

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ✔️

### Proxy

Added the ability to bind an interface as a proxy using the `Spiral\Core\Config\Proxy` configuration. This is useful in cases where a service needs to be used within a specific scope but must be accessible within the container for other services in root or other scopes (so that a service requiring the dependency can be successfully created and used when needed in the correct scope).

```php
use Spiral\Boot\Bootloader\Bootloader;
use Spiral\Core\BinderInterface;
use Spiral\Core\Config\Proxy;
use Spiral\Framework\ScopeName;
use Spiral\Http\PaginationFactory;
use Spiral\Pagination\PaginationProviderInterface;

final class PaginationBootloader extends Bootloader
{
    public function __construct(
        private readonly BinderInterface $binder,
    ) {
    }
    
    public function defineSingletons(): array
    {
        $this->binder
            ->getBinder(ScopeName::Http)
            ->bindSingleton(PaginationProviderInterface::class, PaginationFactory::class);
        
        $this->binder->bind(
            PaginationProviderInterface::class,
            new Proxy(PaginationProviderInterface::class, true)  // <-------
        );

        return [];
    }
}
```

### DeprecationProxy

Similar to **Proxy**, but also allows outputting a deprecation message when attempting to retrieve a dependency from the container. In the example below, we use two bindings, one in scope and one out of scope with `Spiral\Core\Config\DeprecationProxy`. When requesting the interface in scope, we will receive the service, and when requesting it out of scope, we will receive the service and a deprecation message.

```php
use Spiral\Boot\Bootloader\Bootloader;
use Spiral\Core\BinderInterface;
use Spiral\Core\Config\DeprecationProxy;
use Spiral\Framework\ScopeName;
use Spiral\Http\PaginationFactory;
use Spiral\Pagination\PaginationProviderInterface;

final class PaginationBootloader extends Bootloader
{
    public function __construct(
        private readonly BinderInterface $binder,
    ) {
    }

    public function defineSingletons(): array
    {
        $this->binder
            ->getBinder(ScopeName::Http)
            ->bindSingleton(PaginationProviderInterface::class, PaginationFactory::class);

        $this->binder->bind(
            PaginationProviderInterface::class,
            new DeprecationProxy(PaginationProviderInterface::class, true, ScopeName::Http, '4.0') // <----------
        );

        return [];
    }
}
```

Psalm fixed in: #1076

### DIFF
--- a/src/Core/src/Config/DeprecationProxy.php
+++ b/src/Core/src/Config/DeprecationProxy.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Spiral\Core\Config;
+
+final class DeprecationProxy extends Proxy
+{
+    /**
+     * @param class-string $interface
+     */
+    public function __construct(
+        string $interface,
+        bool $singleton = false,
+        private readonly string|\BackedEnum|null $scope = null,
+        private readonly ?string $version = null,
+        private readonly ?string $message = null,
+    ) {
+        if (($scope === null || $version === null) && $message === null) {
+            throw new \InvalidArgumentException('Scope and version or custom message must be provided.');
+        }
+
+        parent::__construct($interface, $singleton);
+    }
+
+    /**
+     * @return class-string
+     */
+    public function getInterface(): string
+    {
+        $message = $this->message ?? \sprintf(
+            'Using `%s` outside of the `%s` scope is deprecated and will be impossible in version %s.',
+            $this->interface,
+            $this->scope,
+            $this->version
+        );
+
+        @trigger_error($message, \E_USER_DEPRECATED);
+
+        return parent::getInterface();
+    }
+}
--- a/src/Core/src/Config/DeprecationProxy.php
+++ b/src/Core/src/Config/DeprecationProxy.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Spiral\Core\Config;
 
+/**
+ * @internal
+ */
 final class DeprecationProxy extends Proxy
 {
     /**
--- a/src/Core/src/Config/Proxy.php
+++ b/src/Core/src/Config/Proxy.php
@@ -15,16 +15,16 @@ class Proxy extends Binding
     ) {
     }
 
+    public function __toString(): string
+    {
+        return \sprintf('Proxy to `%s`', $this->interface);
+    }
+
     /**
      * @return class-string
      */
     public function getInterface(): string
     {
         return $this->interface;
-    }
-
-    public function __toString(): string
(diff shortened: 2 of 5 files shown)


src/Core/src/Config/DeprecationProxy.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Spiral\Core\Config;
+
+final class DeprecationProxy extends Proxy
```
roxblnfk, line 7: Add `@deprecated` annotation?

src/Core/src/Internal/Factory.php
```diff
@@ -194,8 +193,23 @@ private function resolveAlias(
...
+        Config\Proxy $binding,
+        string $alias,
+        Stringable|string|null $context,
+        array $arguments,
+    ): mixed {
+        $result = Proxy::create(new \ReflectionClass($binding->getInterface()), $context, new Attribute\Proxy());
+
+        if ($binding->singleton && $arguments === []) {
+            $this->state->singletons[$alias] = $result;
+        }
```
roxblnfk, line 206:
```suggestion
        if ($binding->singleton) {
            $this->state->singletons[$alias] = $result;
        }
```

src/Core/src/Config/Proxy.php
```diff
@@ -0,0 +1,30 @@
...
+class Proxy extends Binding
+{
+    /**
+     * @param class-string $interface
+     */
+    public function __construct(
+        protected readonly string $interface,
+        public readonly bool $singleton = false,
+    ) {
+    }
```
roxblnfk, line 16: I think it will be nice to check that `$interface` is an interface and thow InvalidArgumentException if not.

src/Core/src/Config/DeprecationProxy.php
msmakouz: Added `@internal`.  This question is open for further discussion.

src/Core/src/Internal/Factory.php
msmakouz: Removed.

src/Core/src/Config/Proxy.php
msmakouz: Added

roxblnfk (approved):

PR #1079, opened 2024-02-16
Added `PaginationProviderInterface` binding in scope

msmakouz:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌

### DIFF
--- a/src/Framework/Http/PaginationFactory.php
+++ b/src/Framework/Http/PaginationFactory.php
@@ -6,8 +6,10 @@ namespace Spiral\Http;
 
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use Spiral\Core\Attribute\Scope;
 use Spiral\Core\Exception\ScopeException;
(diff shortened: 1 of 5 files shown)


src/Framework/Http/PaginationFactory.php
```diff
@@ -6,15 +6,18 @@
...
 use Spiral\Core\FactoryInterface;
+use Spiral\Framework\Spiral;
 use Spiral\Pagination\PaginationProviderInterface;
 use Spiral\Pagination\Paginator;
 use Spiral\Pagination\PaginatorInterface;
 
 /**
  * Paginators factory binded to active request scope in order to select page number.
  */
+#[Scope(Spiral::HttpRequest)]
```
roxblnfk, line 20: > [!NOTE]
> this attribute will be breaking change in the case when the user requests the implementation directly, not via interface

roxblnfk (approved):

PR #1084, opened 2024-02-20
Init Core Introspector

roxblnfk:
Internal feature


msmakouz:
@roxblnfk @butschster Maybe it would be better to add this to spiral/dumper? Because it is only needed for debugging

roxblnfk:
it's internal thing and may be moved/removed/open in the future. Right now it's just a concept

msmakouz (approved):

Issue #1074, opened 2024-02-10
newLine() does not exist in console app


lifewcody:
### Description

protected function alert(string $string): void

in HelpersTrait.php calls the following:

$this->newLine();

Which calls $this->output->newLine($count);

Which returns:

[Error]
Call to undefined method Symfony\Component\Console\Output\ConsoleOutput::newLine()
in vendor/spiral/framework/src/Console/src/Traits/HelpersTrait.php:197

### How To Reproduce

In the console application, set to production/prod and call $this->alert(),

### Additional Info

| Q                   | A
| ------------------- | ---
| Framework Version   | 3.11
| PHP version         | 8.3.1
| Operating system    | Linux

lifewcody:
I should have been more specific; It's only when calling `confirmToProceed()`

But then I'm not sure why alert works but not in `confirmToProceed()`

```
if (!$confirmation->confirmToProceed()) {
            return self::FAILURE;
        }
```

butschster:
Hi @lifewcody, thank you for the bug report! Yes, I was able to reproduce the problem. We will fix it soon.

PR #1085, opened 2024-02-21
Console scopes

butschster:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ✔️
| New feature?  | ✔️
| Issues        | #1074

### DIFF
--- a/src/Console/src/Command.php
+++ b/src/Console/src/Command.php
@@ -97,7 +97,7 @@ abstract class Command extends SymfonyCommand implements EventDispatcherAwareInt
                         ],
                         autowire: true,
                     ),
-                    fn(CommandCoreFactory $factory) => $factory->make(
+                    fn (CommandCoreFactory $factory) => $factory->make(
                         $this->interceptors,
                         $this->eventDispatcher,
                     )->callAction(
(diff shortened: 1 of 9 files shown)


msmakouz (approved):

src/Console/src/Command.php
```diff
@@ -73,27 +73,43 @@ public function setEventDispatcher(EventDispatcherInterface $eventDispatcher): v
     /**
      * Pass execution to "perform" method using container to resolve method dependencies.
      */
-    protected function execute(InputInterface $input, OutputInterface $output): int
+    final protected function execute(InputInterface $input, OutputInterface $output): int
```
roxblnfk, line 76: I think we should use `@final` annotation instead. Because of BC

msmakouz: @roxblnfk Agree, changed to phpdoc

src/Console/src/CommandCoreFactory.php
```diff
@@ -0,0 +1,36 @@
...
+
+#[Scope('console.command')]
+final class CommandCoreFactory
+{
+    public function __construct(
+        private readonly ContainerInterface $container,
+    ) {
+    }
+
+    public function make(array $interceptors, ?EventDispatcherInterface $eventDispatcher = null): CoreInterface
```
roxblnfk, line 22: Add $interceptor php doc with type

roxblnfk (approved):

PR #1098, opened 2024-04-02
Fixed bug in directive parser

butschster:
Fixed bug in directive parser where values containing zero character, such as '150' in '@image("blog", "test.png", "150|250", "webp")', were not processed correctly.

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️


msmakouz (approved):

src/Stempler/src/Parser/Syntax/DynamicSyntax.php
```diff
@@ -82,7 +82,7 @@ private function fetchValues(string $body): array
         while ($n = $src->peak()) {
             if (\in_array($n, ['"', '"'])) {
```
roxblnfk, line 83: Fix quotes

roxblnfk (approved):

PR #1104, opened 2024-04-30
Integrate Container Scopes

roxblnfk:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ✔️ 
| Issues        | #
| Docs PR       |


composer.json
```diff
@@ -136,7 +136,7 @@
         "rector/rector": "0.18.1",
         "spiral/code-style": "^1.1",
         "spiral/nyholm-bridge": "^1.2",
-        "spiral/testing": "^2.7",
+        "spiral/testing": "dev-feature/scopes as 2.8.0",
```
roxblnfk, line 139: @msmakouz what do we need to release spiral/testing 2.8.0?

msmakouz: In the **feature/scopes** branch, the `TestScope` attribute has been added.

Before the release, it is necessary to check the `TestScope` attribute together with the `Config` attribute. There were issues with this, the configuration was being changed twice, causing an exception.

msmakouz: Released.

PR #1117, opened 2024-07-29
Update interceptors

roxblnfk:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ✔️ 

Add CallableHandler that just calls callables
Fix interceptor HandlerInterface: use CallContextInterface instead of the CallContext class

### DIFF
--- a/src/Hmvc/src/Interceptors/Handler/ReflectionHandler.php
+++ b/src/Hmvc/src/Interceptors/Handler/ReflectionHandler.php
@@ -7,7 +7,6 @@ namespace Spiral\Interceptors\Handler;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\ContainerInterface;
 use Spiral\Core\ResolverInterface;
-use Spiral\Interceptors\Context\CallContext;
 use Spiral\Interceptors\Context\CallContextInterface;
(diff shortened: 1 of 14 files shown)


src/Hmvc/src/Interceptors/Handler/ReflectionHandler.php
```diff
@@ -30,7 +31,7 @@ public function __construct(
      * @psalm-assert non-empty-string $action
      * @throws \Throwable
      */
-    public function handle(CallContext $context): mixed
+    public function handle(CallContextInterface $context): mixed
```
msmakouz, line 34: Perhaps it is also necessary to replace CallContext with CallContextInterface here (line 78)?
```php
protected function resolveArguments(\ReflectionFunctionAbstract $method, CallContext $context): array
```

msmakouz (changes requested):

src/Hmvc/src/Interceptors/Handler/ReflectionHandler.php
roxblnfk: Thanks. Fixed.

butschster (approved):

PR #1119, opened 2024-08-08
[psalm fix level 4] TracerFactoryInterface - not internal

gam6itko:
| Q             | A
| ------------- | ---
| Bugfix?       | ❌
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | #... 

Remove internal annotation from:
- TracerFactoryInterface
- TracerInterface

Psalm level 4 triggers error
```php

use Spiral\Telemetry\TracerFactoryInterface;

final class TelemetryDecorator implements ResponseDataHandlerInterface
{
    public function __construct(
        private readonly TracerFactoryInterface       $tracerFactory,
        private readonly ResponseDataHandlerInterface $inner,
    ) {
    }

    public function handle(array $responseData, array $options = []): array
    {     
        return $this->tracerFactory
            ->make()
            ->trace(
                name: sprintf('ResponseDataHandler [%s]', get_class($this->inner)),
                callback: fn () => $this->inner->handle($responseData, $options),
                attributes: [
                    'itemsCount' => count($responseData),
                    'options' => $options,
                ]
            );
    }
}
```

```
ERROR: InternalMethod
at /home/gam6itko/TelemetryDecorator.php:23:15
The method Spiral\Telemetry\TracerFactoryInterface::make is internal to Spiral but called from TelemetryDecorator::handle (see https://psalm.dev/175)
            ->make()


ERROR: InternalMethod
at /home/gam6itko/TelemetryDecorator.php:24:15
The method Spiral\Telemetry\TracerInterface::trace is internal to Spiral but called from TelemetryDecorator::handle (see https://psalm.dev/175)
            ->trace(

```

The usage example.
https://spiral.dev/docs/advanced-telemetry/current/en#create-trace-from-context

### DIFF
--- a/src/Telemetry/src/LogTracerFactory.php
+++ b/src/Telemetry/src/LogTracerFactory.php
@@ -11,8 +11,6 @@ use Spiral\Core\ScopeInterface;
 use Spiral\Logger\LogsInterface;
 
 /**
- * The component is under development.
- *
  * Something may be changed in the future. We will stable it soon.
  * Feedback is welcome {@link https://github.com/spiral/framework/discussions/822}.
  */
--- a/src/Telemetry/src/LogTracerFactory.php
+++ b/src/Telemetry/src/LogTracerFactory.php
@@ -11,7 +11,8 @@ use Spiral\Core\ScopeInterface;
 use Spiral\Logger\LogsInterface;
 
 /**
- * @internal The component is under development.
+ * The component is under development.
+ *
  * Something may be changed in the future. We will stable it soon.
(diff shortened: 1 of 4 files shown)


roxblnfk:
If we make these classes non-internal, then the message "The component is under development" needs to be removed.

@butschster is it still under development?

butschster:
> If we make these classes non-internal, then the message "The component is under development" needs to be removed.
> 
> @butschster is it still under development?

Nope, it's all done now. Also, let's delete the other part of the line "The component is under development" as well.

Cheers!

butschster (approved):

PR #1131, opened 2024-09-03
[Rector] Enable RemoveUnusedPublicMethodParameterRector

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

This is part of dead code set list, remove unused public method parameter when class is final.


samsonasik:
Ready to merge 👍

roxblnfk (commented):
We can't change public function signature in this way because it might be used in the userland with named arguments.

I think it's time to create a 4.x branch and start cleaning up unnecessary parameters and other deprecations in it. This aligns with our roadmap.

I will move your PR to 4.x once everything is prepared.

PR #1137, opened 2024-09-04
Fix concurrent writing and reading on workers boot

wapmorgan:
Set a shared lock on a file in Memoty::loadData() to prevent reading an empty file while writing it in concurrent worker/process

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ❌
| Issues        | Can get an empty file when `file_put_contents` in `\Spiral\Boot\Memory::saveData=>\Spiral\Files\Files::write` created, but not completed write to file
| Docs PR       | spiral/docs

Got this problem when 2 jobs consumers concurrently started writing/reading runtime/cache/cycle.php file. First consumer writing schema to the file, second consumers find out that files exists and read it, but it still was empty = so read empty schema.

### DIFF
--- a/src/Boot/src/Memory.php
+++ b/src/Boot/src/Memory.php
@@ -35,7 +35,14 @@ final class Memory implements MemoryInterface
         }
 
         try {
-            return include($filename);
+            $fp = fopen($filename, 'r');
+            if (!flock($fp, LOCK_SH | LOCK_NB)) {
+                return null;
+            }
+            $data = include($filename);
+            flock($fp, LOCK_UN);
+            fclose($fp);
+            return $data;
         } catch (\Throwable) {
             return null;
         }
--- a/src/Boot/src/Memory.php
+++ b/src/Boot/src/Memory.php
@@ -1,4 +1,4 @@
-<?php
+e<?php
 
 declare(strict_types=1);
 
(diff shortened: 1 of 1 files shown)


wapmorgan:
Also another options is to remove `LOCK_NB` and remain only `LOCK_SH` to force current thread to wait until it will be able to make shared lock.
In current MR code current thread will build schema again and write it again to file.

wolfy-j (approved):

roxblnfk (approved):

roxblnfk:
@wapmorgan hi, the previous commit was OK. Could you revert the last commit?

src/Boot/src/Memory.php
```diff
@@ -1,4 +1,4 @@
-<?php
+e<?php
```
roxblnfk, line 1: There is still an extra symbol

roxblnfk, line 1:
```suggestion
<?php
```

src/Boot/src/Memory.php
```diff
@@ -35,7 +35,14 @@ public function loadData(string $section, string &$filename = null): mixed
...
 
         try {
-            return include($filename);
+            $fp = fopen($filename, 'r');
+            if (!flock($fp, LOCK_SH | LOCK_NB)) {
+                return null;
+            }
+            $data = include($filename);
+            flock($fp, LOCK_UN);
+            fclose($fp);
```
roxblnfk, line 44:
```suggestion
            $fp = \fopen($filename, 'r');
            if (!\flock($fp, \LOCK_SH | \LOCK_NB)) {
                return null;
            }
            $data = include($filename);
            \flock($fp, \LOCK_UN);
            \fclose($fp);
```

roxblnfk (approved):

PR #1144, opened 2024-09-10
Enable Auto import Rector config

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

- [x] Enable auto import with Rector
- [x] Sort use statement via php-cs-fixer.


samsonasik:
Ready to merge 👍

samsonasik:
that's not supported right now, as auto import is really complex process, we check it whether it is conflicted as used by alias as well.

samsonasik:
Let's close for now then ;)

PR #1149, opened 2024-09-20
Container scope related fixes

roxblnfk:
- Fix DebugBootloader: unresolved collectors don't break state populating
- Router: Fix opening custom scopes in middleware
- More compatible Handler resolving in RouterBootloader

### DIFF
--- a/src/Framework/Bootloader/Http/RouterBootloader.php
+++ b/src/Framework/Bootloader/Http/RouterBootloader.php
@@ -107,7 +107,7 @@ final class RouterBootloader extends Bootloader
         );
     }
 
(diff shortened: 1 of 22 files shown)


src/Framework/Bootloader/Http/RouterBootloader.php
```diff
@@ -107,6 +107,13 @@ private function router(
         );
     }
 
+    private function handler(?CoreInterface $core = null, ContainerInterface $container): HandlerInterface
```
rauanmayemir, line 110: This will fix a fatal error:

```suggestion
    private function handler(
        #[Proxy] ContainerInterface $container,
        ?CoreInterface $core = null,
    ): HandlerInterface {
```

PR #1151, opened 2024-09-29
fix: boot resource leakage

leon0399:
| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | N/A 
| Docs PR       | N/A 

#1137 introduced resource leakage by not freeing handles in some cases. This led to endless loop while trying to loadData or saveData

This PR should fix this bug by always attempting to close the resource

### DIFF
--- a/src/Boot/src/Memory.php
+++ b/src/Boot/src/Memory.php
@@ -24,9 +24,9 @@ final class Memory implements MemoryInterface
     }
 
     /**
-     * @param string $filename Cache filename.
+     * @param non-empty-string|null $filename Cache filename.
      */
-    public function loadData(string $section, string &$filename = null): mixed
+    public function loadData(string $section, ?string &$filename = null): mixed
     {
         $filename = $this->getFilename($section);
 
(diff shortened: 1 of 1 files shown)


roxblnfk (approved):

src/Boot/src/Memory.php
```diff
@@ -36,16 +36,27 @@ public function loadData(string $section, string &$filename = null): mixed
...
             $data = include($filename);
+
             \flock($fp, \LOCK_UN);
-            \fclose($fp);
-            return $data;
         } catch (\Throwable) {
             return null;
+        } finally {
+            if (isset($fp)) {
+                \fclose($fp);
```
roxblnfk, line 55: Looks like `$fp` might be `false` there

leon0399: Here's a Psalm report from the previous run with a check: 

https://github.com/spiral/framework/actions/runs/11093746018/job/30821292030

roxblnfk:
LGTM

PR #1159, opened 2024-10-18
Bump to Rector 1.2.8

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

Bump to just released Rector 1.2.6 and re-run it https://github.com/rectorphp/rector/releases/tag/1.2.8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -142,7 +142,7 @@
         "phpunit/phpunit": "10.5.3",
         "ramsey/collection": "^1.2",
         "ramsey/uuid": "^4.2.3",
-        "rector/rector": "1.2.8",
+        "rector/rector": "~1.2.8",
         "spiral/code-style": "^1.1",
         "spiral/nyholm-bridge": "^1.2",
         "spiral/testing": "^2.8",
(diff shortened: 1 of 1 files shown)


samsonasik:
Ready to merge 👍

composer.json
```diff
@@ -142,7 +142,7 @@
         "phpunit/phpunit": "10.5.3",
         "ramsey/collection": "^1.2",
         "ramsey/uuid": "^4.2.3",
-        "rector/rector": "1.2.6",
+        "rector/rector": "1.2.8",
```
roxblnfk, line 145:
```suggestion
        "rector/rector": "~1.2.8",
```

Maybe we should just lock the version this way? :)

samsonasik: we are follow phpstan change :), sometime, it break, for personal repo, that's fine, but for repo with many contributors, I think pinned version is better, better to use commit composer.lock as well imo, and use renovate or dependabot to update so when it break, it can directly be executed/skipped by maintainer :)

samsonasik: I updated to use ~1.2.8 anyway :) https://github.com/spiral/framework/pull/1159/commits/59bced6285886eff8a436825c3bb1dc38c770927

samsonasik:
Ready to merge 👍

roxblnfk (approved):

PR #1163, opened 2024-11-08
Update to use PHPUnit 10 syntax

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

Updated to latest PHPUnit 10 syntax version based on defined in composer.json with enable phpunit prepared set.


samsonasik:
Ready to merge 👍

roxblnfk (approved):

PR #1180, opened 2024-12-29
refactor: add closure void return type via rector type declaration set level 0

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

Apply `AddClosureVoidReturnTypeWhereNoReturnRector` to add closure return type void if there is no return


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1186, opened 2025-01-04
refactor: enable phpunit code quality set for rector

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

Apply `phpunitCodeQuality` prepared set into rector config to apply refactor on tests.


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk:
Is there a discussion on why 'this' is more preferable than 'self'?

samsonasik:
see this reasoning:

- https://github.com/rectorphp/rector-phpunit/issues/413#issuecomment-2528829737

I can skip the rule if you want it to be skipped

samsonasik:
@roxblnfk there are also incosistent usage over code base which some of them uses `$this`, and other uses `self::`, I think using `$this` is fine also for consistency.

roxblnfk:
It might be inconvenient to constantly correct `self` to `$this`, since autocomplete always inserts `self`:

ani

That's why I think it's better not to force `$this` or use `PreferPHPUnitSelfCallRector` instead of `PreferPHPUnitThisCallRector`.

roxblnfk:
> @roxblnfk there are also incosistent usage over code base which some of them uses `$this`, and other uses `self::`, I think using `$this` is fine also for consistency.
> 
> see [#1186 (files)](https://github.com/spiral/framework/pull/1186/files#diff-739540189da5dc9e63300d668f22ea9c068decb1c14a1d304a45aa8ea1df3dbd)

In the last components we used `self::` :)

samsonasik:
@roxblnfk ok, I updated to use `self::` via `PreferPHPUnitSelfCallRector` : https://github.com/spiral/framework/pull/1186/commits/8772eeb79a0326a1d1f451a9c6759506fa178002

samsonasik:
@roxblnfk ready to merge 👍

src/Console/tests/AttributeTest.php
```diff
@@ -62,9 +53,6 @@ public function testCommandWithSymfonyAttribute(): void
             WithSymfonyAttributeCommand::class
         ]));
 
-        $this->assertSame(
-            'Some description text|attribute-with-sf-command-attr',
-            $core->run(command: 'attribute-with-sf-command-attr')->getOutput()->fetch()
-        );
+        self::assertSame('Some description text|attribute-with-sf-command-attr', $core->run(command: 'attribute-with-sf-command-attr')->getOutput()->fetch());
```
roxblnfk, line 56: Why does it squash multiline assertions into inline ones?

samsonasik: It rely on AST printing, I can rollback new line manually if needed.

Any coding style I can just run it? I tried exising cs fixer and it seems change many unrelated change

roxblnfk: > It rely on AST printing

Ah, okay. Let's leave it as is. I'll set up CS Fixer in the future.

roxblnfk (approved):

Issue #1154, opened 2024-10-04
Stempler Import Directory does not work when mixed with HTML


enyancc:
### Description
Adding `<use:dir/>` tag within template that contains HTML would try to resolve regular tags as templates.

### How To Reproduce

**views/home.dark.php**
```html
<extends:layout.base title="[[The PHP Framework for future Innovators]]"/>
<use:dir dir="partials" ns="partials"/>
<stack:push name="styles">
    <link rel="stylesheet" href="/styles/welcome.css"/>
</stack:push>

<define:body>
    <h1 class="main-title">The PHP Framework <span>for future Innovators</span></h1>
    <partials:router-link />
</define:body>
```

**views/partials/router-link.dark.php**
```html
<a href="test.php">test</a>
```

throws exception "Unable to resolve import `link`"

```html
<extends:layout.base title="[[The PHP Framework for future Innovators]]"/>
<use:dir dir="partials" ns="partials"/>
<stack:push name="styles">
    <link rel="stylesheet" href="/styles/welcome.css"/> <!------ HERE  ->
</stack:push>
 
<define:body>
    <h1 class="main-title">The PHP Framework <span>for future Innovators</span></h1>
    <partials:router-link/>
</define:body>

```

### Additional Info

| Q                   | A
| ------------------- | ---
| Framework Version   | 3.14.5 
| PHP version         | 8.3.4 
| Operating system    | Linux

butschster:
Hi @enyancc
Thx for the bug report!

PR #1191, opened 2025-01-05
Fix directory import in stempler component

butschster:
see #1154

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌ 
| New feature?  | ❌ 
| Issues        | #1154

### DIFF
--- a/src/Stempler/src/Node/Template.php
+++ b/src/Stempler/src/Node/Template.php
@@ -23,18 +23,12 @@ final class Template implements NodeInterface, AttributedInterface
      * @param list<TNode> $nodes
      */
     public function __construct(
-        public array $nodes = []
-    ) {
-    }
+        public array $nodes = [],
+    ) {}
 
     public function setContext(?Context $context = null): void
     {
         $this->context = $context;
-    }
-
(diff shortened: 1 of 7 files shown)


src/Stempler/src/Node/Template.php
```diff
@@ -23,20 +23,14 @@ final class Template implements NodeInterface, AttributedInterface
...
-    }
+        public array $nodes = [],
+    ) {}
 
     public function setContext(?Context $context = null): void
     {
         $this->context = $context;
     }
 
-    public function getContext(): ?Context
```
butschster, line 35: This method is presented in Context trait

src/Stempler/src/Node/Template.php
```diff
@@ -23,20 +23,14 @@ final class Template implements NodeInterface, AttributedInterface
...
-    {
-        return $this->context;
-    }
-
```
roxblnfk, line 39: Looks like BC here

roxblnfk (approved):

roxblnfk (approved):

Issue #1162, opened 2024-10-26
[spiral/router] Required parameters are not required for uri generation


MartkCz:
### Description

When a parameter is not passed, even though it is required in a route, it is ignored without raising an error.

### How To Reproduce

```php
$routes->add('articles.list', '/articles')
    ->callable(fn () => new Response());
$routes->add('articles.detail', '/articles/<id>')
    ->callable(fn () => new Response());

var_dump($router->uri('articles.detail')->getPath()); // == /articles - Better solution is exception or warning
var_dump($router->uri('articles.list')->getPath()); // == /articles
var_dump($router->uri('articles.detail', ['id' => 1])->getPath()); // == /articles/1 
```

butschster:
Hey, @MartkCz!!

Thanks for reporting this issue—I've checked it out, and it does seem like a bug!

> @butschster hi Could you comment the issue? Is it a bug or a feature?

From what I see, the route pattern should be parsed to identify required parameters, and the URI resolver should validate that all necessary data is provided.

PR #1192, opened 2025-01-06
fix: add strict mode for UriHandler

butschster:
Strict mode ensures all required URI segments are validated. If any are missing, an exception is thrown. 

To enable strict mode, request `Spiral\Router\UriHandler` from the container in `RoutersBootloader` and set strict mode:

```php
$handler = $container->get(\Spiral\Router\UriHandler::class);
$handler->setStrict(true);
```

| Q             | A
| ------------- | ---
| Bugfix?       | ✔️
| Breaks BC?    | ❌
| New feature?  | ✔️
| Issues        | #1162

### DIFF
--- a/src/Router/src/UriHandler.php
+++ b/src/Router/src/UriHandler.php
@@ -45,7 +45,10 @@ final class UriHandler
     private string $basePath = '/';
     private ?string $compiled = null;
     private ?string $template = null;
+
     private array $options = [];
+    private array $requiredOptions = [];
+    private bool $strict = false;
 
     private \Closure $pathSegmentEncoder;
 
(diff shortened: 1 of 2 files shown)


src/Router/src/UriHandler.php
```diff
@@ -201,15 +210,33 @@ public function uri(iterable $parameters = [], array $defaults = []): UriInterfa
...
             if (empty($parameters[$key])) {
-                throw new UriHandlerException(\sprintf('Unable to generate Uri, parameter `%s` is missing', $key));
+                $missingParameters[] = $key;
             }
         }
 
+        if ($missingParameters !== []) {
+            throw new UriHandlerException(
+                \sprintf(
+                    \count($missingParameters) === 1
```
butschster, line 232: Refactor error message

roxblnfk (approved):

PR #1196, opened 2025-01-06
refactor: use instanceof over null/empty comparison with enable instanceOf rector set

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

- immediate know what type object to be used via instanceof
- flip ternary negation instanceof to check truthy first.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -120,7 +120,7 @@ return RectorConfig::configure()
         PreferPHPUnitThisCallRector::class,
     ])
     ->withPhpSets(php81: true)
-    ->withPreparedSets(deadCode: true, phpunitCodeQuality: true)
+    ->withPreparedSets(deadCode: true, phpunitCodeQuality: true, instanceOf: true)
     ->withComposerBased(phpunit: true)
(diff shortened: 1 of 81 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk:
@samsonasik hi.
This change seems to degrade performance. A null check is faster than accessing reflection, isn't it?

samsonasik:
The target of this PR is to make more exclusive check, instead of inconsistence check, there are `empty()`, or `null`, and there are other `instanceof`.

The benefit is we know what type to check otherwise, with `null` or `empty()` only, we don't know what actually type to check, that's just guessing.

with more exclusive check, we know the exact type to check/use.

roxblnfk:
We rely on the fact that if a property can only contain two types of values: `null` or an instance of X. In this case, it's obvious that if the value is not `null`, it is exactly an instance of X.

I know how to hacks and brake types, but we don't need to try to catch that in the code.

samsonasik:
Ok, let's close then ;)

PR #1235, opened 2025-09-10
set QueueInterface $payload parameter type to mixed

algirdasci:
## What was changed

Changed `QueueInterface` `$payload` parameter from `array` to `mixed`.

## Why?

All interface implementations uses `mixed` payload where providing serializable objects are possible. Using `QueueInterface` as type, triggers phpstan errors of incompatible types:

Example:
```php
class CallbackService
{
    public function __construct(
        // flow is in-memory roadrunner driver (Spiral\RoadRunnerBridge\Queue\Queue)
        private readonly QueueInterface $flow,
    ) {
    }
    
    public function send(): void
    {
        // CallbackJob is DTO which can be serialized
        $job = new CallbackJob();
    
        // call below will trigger phpstan error:
        // phpstan: Parameter #2 $payload of method Spiral\Queue\QueueInterface::push() expects array, CallbackJob given.
        $this->flow->push(CallbackJob, $job);
    }
}
```

### DIFF
--- a/src/Queue/src/QueueInterface.php
+++ b/src/Queue/src/QueueInterface.php
@@ -9,5 +9,5 @@ interface QueueInterface
     /**
      * @param string|class-string<HandlerInterface> $name
      */
-    public function push(string $name, array $payload = [], ?OptionsInterface $options = null): string;
+    public function push(string $name, mixed $payload = [], ?OptionsInterface $options = null): string;
 }


roxblnfk:
Expanding the parameter type in the interface is a breaking change for backward compatibility. We cannot accept this change in the current version.

However, we can consider declaring a broader type with annotations.

PR #1260, opened 2026-06-05
Add StructArmed to QA

samsonasik:
## What was changed

This PR adds StructArmed to github workflow static analysis for QA.

https://github.com/boundwize/structarmed

StructArmed is a configurable PHP architecture guard, that we can define layers and rules, then keep them enforced. 

For starter, this PR uses the PSR4 preset, and already found violations that fixed and included in this PR.

Since this repo support php 8.1 as well, this only add on github workflow on php 8.2 👍

On more practical use case of how set the config layer -> ruleset for more complex use, you can take a look on how cakephp and codeigniter4 uses it :)

- https://github.com/cakephp/cakephp/blob/5.next/structarmed.php
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/structarmed.php

## Why?

This adds an automated architecture guard, and for starter, add psr4 preset for the starter only to give quick entrance.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,4 @@ phpcs.xml              export-ignore
 phpunit.xml            export-ignore
 psalm.xml              export-ignore
 rector.php             export-ignore
+structarmed.php        export-ignore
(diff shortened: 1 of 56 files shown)


samsonasik:
@roxblnfk ready to review/merge 👍

roxblnfk (approved):
The tool looks interesting. Are you planning to expose an agent skill (e.g. a SKILL.md) for it, so coding agents can configure and run it without a human reading the whole README?

samsonasik:
@roxblnfk I probably not going to add `SKILL.md`, at least for now.

PR #1261, opened 2026-06-05
Enable FinalizeTestCaseClassRector

samsonasik:
## What was changed

Finalize test classes when possible.

## Why?

Test case classes should not be extended.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1262, opened 2026-06-06
Update composer audit to run without dev dependencies and run composer cs:fix

samsonasik:
Change composer audit command to exclude dev dependencies.

## What was changed

Use of `--no-dev` for run composer audit and run cs fix.

## Why?

Phpunit is not used for production as dev dependency.

This PR try to make security audit green.

Also run cs fix per master notice due to class is final so static can be updated to self.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
 
       # Execution
       - name: Audit dependencies
-        run: composer audit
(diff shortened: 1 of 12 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1263, opened 2026-06-06
Enable CreateStubOverCreateMockArgRector

samsonasik:
## What was changed

Use of createStub() when possible. 

## Why?

Use createStub() over createMock() when used as argument or array value and does not add any mock requirements.

## Checklist

- Tested
    - [x] Tested manually
    - [ ] Unit tests added

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -130,7 +130,6 @@ return RectorConfig::configure()
         ],
 
         // to be enabled later for easier to review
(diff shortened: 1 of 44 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1266, opened 2026-06-10
Enable DeclareStrictTypesTestsRector

samsonasik:
## What was changed

This PR enable `DeclareStrictTypesTestsRector`

## Why?

Test class file which should be safe to use `declare(strict_types=1)` so this PR adds it.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -130,7 +130,6 @@ return RectorConfig::configure()
         ],
 
         // to be enabled later for easier to review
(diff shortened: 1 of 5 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1268, opened 2026-06-10
Enable RemoveNullArgOnNullDefaultParamRector

samsonasik:
## What was changed

Remove default null argument, where null is already a default param value

## Why?

If null argument is equal with default param, no need to add it as argument.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1271, opened 2026-06-13
Enable AddInstanceofAssertForNullableInstanceRector

samsonasik:
## What was changed

Add explicit instance assert between nullable object assign and method call on nullable object.

## Why?

Adding an explicit instance assertion gives static analysis a precise type boundary before the method is called.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -135,8 +135,6 @@ return RectorConfig::configure()
         ],
 
         // to be enabled later for easier to review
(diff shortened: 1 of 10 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1275, opened 2026-06-16
Enable NarrowWideUnionReturnTypeRector

samsonasik:
## What was changed

Narrow overly wide union return type declaration if possible.

## Why?

Narrowing the union type better reflects the real behavior, improves type inference, and reduces unnecessary downstream checks without changing runtime behavior.

This only apply if:

- class is final class 
- class is not final, but method is private

so this is safe change.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -134,8 +134,13 @@ return RectorConfig::configure()
             __DIR__ . '/src/Reactor/tests/Partial/MethodTest.php',
         ],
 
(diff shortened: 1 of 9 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1276, opened 2026-06-16
Enable Rector TypeCoverageLevel 10

samsonasik:
## What was changed

Applied rule: `StringReturnTypeFromStrictScalarReturnsRector`.

## Why?

We can add string return type based on returned string scalar values.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1279, opened 2026-06-20
Enable Rector TypeCoverageLevel 14

samsonasik:
## What was changed

Applied rule: `NumericReturnTypeFromStrictReturnsRector`.

## Why?

We can add int/float return type based on strict typed returns.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1280, opened 2026-06-20
Enable Rector TypeCoverageLevel 16

samsonasik:
## What was changed

Applied rule: `ReturnTypeFromReturnDirectArrayRector`.

## Why?

We can add return type from return direct array

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1281, opened 2026-06-20
Enable Rector TypeCoverageLevel 18

samsonasik:
## What was changed

Applied rule: `ReturnTypeFromReturnNewRector`.

## Why?

We can add return type to function like with return new.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -33,9 +33,6 @@
     </InvalidArgument>
   </file>
   <file src="src/Core/src/Internal/Actor.php">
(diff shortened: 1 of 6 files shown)


samsonasik:
@roxblnfk ready to merge 👍

samsonasik:
It seems cause generic notice on psalm, I will looking into it.

samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1282, opened 2026-06-21
Enable Rector TypeCoverageLevel 21

samsonasik:
## What was changed

Applied rule: `AddVoidReturnTypeWhereNoReturnRector`.

## Why?

We can add return type void to function like without any return.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1283, opened 2026-06-21
Enable Rector TypeCoverageLevel 23

samsonasik:
## What was changed

Applied rule: `ReturnNullableTypeRector`.

## Why?

We can add  `?nullabl`e type to class methods and functions

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1286, opened 2026-06-29
Bump structarmed to ^0.14 and use new expanded +LayerName in the config

samsonasik:
## What was changed

This PR bump to latest ^0.14 and apply this config shorthand in `structarmed.php`.

## Why?

Since StructArmed 0.11.0, already brings support for the +LayerName shorthand in ruleset definition.

implemented on:

- https://github.com/boundwize/structarmed/pull/124

Each `+LayerName` entry expands to that layer itself plus all layers it is allowed to depend on.

Ref https://boundwize.github.io/structarmed/layers-and-rulesets/#inheriting-allowed-layers

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1289, opened 2026-07-06
Enable RemoveUselessUnionReturnDocblockRector

samsonasik:
## What was changed

This PR enable `RemoveUselessUnionReturnDocblockRector` that remove useless union docblock, applied changes are on trait.

## Why?

If return type is already object type on trait and return is cloned of `$this`, union to `$this` is not needed.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1294, opened 2026-07-20
Bump to StructArmed ^0.15 and enable MustBeFinalRule on tests

samsonasik:
## What was changed

StructArmed 0.15.0 introduce `ExtendedClassAwareRuleInterface` that makes `MustBeFinalRule` safely be used to skip if there is class that extend target class.

This PR bump StructArmed to `^0.15` and apply the `MustBeFinalRule` with single command to fix.

```bash
vendor/bin/structarmed analyze --fix
```

## Why?

Make use of `final` class when possible.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to review/merge 👍

roxblnfk (approved):

PR #1295, opened 2026-07-28
Bump Rector to ~2.5.8 and skip new rules

samsonasik:
## What was changed

This PR bump Rector to ~2.5.8 and skip new rules

## Why?

The skipped rules are because currently still buggy or needs some separate PR to ease for review, eg the `RemoveDefaultValueFromAssignedPropertyRector` is buggy on non-final class on protected property, fix is on the way on rector side:

- https://github.com/rectorphp/rector-src/pull/8214 

Here skip for now.

## Checklist

- Tested
    - [x] Tested manually

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -142,7 +142,7 @@
         "mockery/mockery": "^1.6.12",
         "phpunit/phpunit": "^10.5.41",
         "ramsey/collection": "^1.3",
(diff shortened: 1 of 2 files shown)


samsonasik:
@roxblnfk ready to merge 👍

roxblnfk (approved):

PR #1297, opened 2026-08-04
Bump Rector to ~2.6.1 and re-run it

samsonasik:
## What was changed

This PR bump Rector to ~2.6.1 and apply rector re-run.

## Why?

Changed parts:

- fix invalid docblock due to allowed directory value in array docblock (manually apply due to rector run apply invalid change to change return type instead), the fix is in the way:

- https://github.com/rectorphp/rector-src/pull/8282

- add return type on willReturnCallback() that apply `ClosureReturnTypeRector`.

## Checklist

- Tested
    - [x] Tested manually

PR #1298, opened 2026-08-04
Fix unit test error on latest php 8.4 and php 8.5 build

samsonasik:
## What was changed

Related by latest build error 

https://github.com/spiral/framework/actions/runs/30915207986/job/92011479490?pr=1297#step:11:84

## Why?

To fix latest build that seems due to latest php release.

## Checklist

- Tested
    - [x] Tested manually


samsonasik:
@roxblnfk ready to merge 👍. Rector notice is handled on PR:

- https://github.com/spiral/framework/pull/1297

roxblnfk:
Thanks! I think it's better to fix the tests rather than the renderer itself: both string representations are valid PHP expressions that evaluate to the same value, so the generated proxy code was never broken — only the literal assertion in the test. I've pushed an update that makes the test accept both variants via `assertContains`, without maintaining a hand-rolled string-literal normalizer in the production code.

samsonasik:
@roxblnfk yes 👍

PR #1299, opened 2026-08-05
[Stempler] Encode on* attribute output as a JavaScript literal

roxblnfk:
## What was changed

Context-aware escaping now encodes both JavaScript contexts as JavaScript literals.

**`on*` attributes** were encoded with `htmlspecialchars()` and wrapped in literal `&quot;`. They are now encoded with the same `JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT` flags already used for `<script>`, with the delimiters HTML-encoded afterwards.

**Invalid UTF-8** no longer collapses the value: `JSON_INVALID_UTF8_SUBSTITUTE` was added to the `on*` branch (first commit) and to the `<script>` branch (second commit).

Three existing assertions on compiled PHP were updated, and four cases were added to `DynamicToPHPTest`.

## Why?

An `on*` attribute value is JavaScript delivered inside HTML, so the browser HTML-decodes it *before* the JS engine parses it. HTML-entity encoding therefore does not hold: `&quot;` and `&#039;` decode back into real quotes, and a quote in the value ends the JavaScript string literal instead of remaining part of it. The `JSON_HEX_*` flags leave no HTML-special character inside the literal, so the value survives both decoding passes — and the outer `htmlspecialchars()` is left with nothing to encode except the surrounding quotes.

This branch is reached for event handlers exclusively, which is why the fix is scoped to them: `HTMLSyntax` marks only `on*` and `style` attribute values as `Verbatim`, and `style` is excluded by the branch condition. Ordinary attributes (`title`, `href`, `data-*`) keep the plain `htmlspecialchars()` filter, where HTML-entity encoding is the correct choice.

Separately, `json_encode()` returns `false` on malformed UTF-8, so the compiled template echoed an empty string and the value vanished from the emitted JavaScript, silently changing the arity of the surrounding call. `JSON_INVALID_UTF8_SUBSTITUTE` mirrors the `ENT_SUBSTITUTE` the default filter has always used.

Rendered output is unchanged for values that needed no encoding — `{{ "hello world" }}` in an `onclick` still renders `alert(&quot;hello world&quot;)` — so only the compiled PHP differs.

## Checklist

- Closes #
- Tested
    - [x] Tested manually
    - [x] Unit tests added

Manual testing covered both branches across strings, quotes, apostrophes, integers, floats, `null`, booleans, lists, associative arrays, non-ASCII text, and malformed UTF-8, comparing the rendered attribute against what a browser yields after HTML-entity decoding. A byte-by-byte sweep confirmed that no HTML-special character survives `json_encode()` with these flags, so the outer `htmlspecialchars()` cannot double-encode: values round-trip back to the original through `html_entity_decode()` + `json_decode()`, including already-encoded input such as `&amp;`.

## Documentation

The "Context-Aware escaping" section of `docs/en/views/stempler.md` documents only `<script>`; it is worth extending with the `on*` case now that both contexts use JavaScript encoding.

### DIFF
--- a/src/Stempler/src/Transform/Finalizer/DynamicToPHP.php
+++ b/src/Stempler/src/Transform/Finalizer/DynamicToPHP.php
@@ -133,7 +133,7 @@ final class DynamicToPHP implements VisitorInterface
             return \sprintf(
                 'json_encode(%s, %s, %s)',
                 '%s',
-                'JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT',
+                'JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT | JSON_INVALID_UTF8_SUBSTITUTE',
                 '512',
             );
         }
--- a/src/Stempler/src/Transform/Finalizer/DynamicToPHP.php
+++ b/src/Stempler/src/Transform/Finalizer/DynamicToPHP.php
@@ -138,9 +138,19 @@ final class DynamicToPHP implements VisitorInterface
             );
         }
 
-        // in on* and other attributes
+        // php {{ }} in on* event handlers. The browser HTML-decodes the attribute value before the JS engine
+        // parses it, so the value must survive both decodings: encode it as a JavaScript literal first, then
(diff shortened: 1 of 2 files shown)


roxblnfk:
@yoskaldyr you might be interested this fix

PR #1302, opened 2026-08-21
Bind MailQueue explicitly in MailerBootloader

gam6itko:
## What

`MailerBootloader` bound `Spiral\Mailer\MailerInterface` with an inline closure inside `boot()`. This PR moves the binding into the singleton map and adds an explicit `MailQueue::class` binding backed by a dedicated factory method:

```php
public function defineSingletons(): array
{
    return [
        // ...
        MailQueue::class => [self::class, 'initMailQueue'],
        MailerInterface::class => MailQueue::class,
    ];
}
```

While here, the `DEPENDENCIES` / `SINGLETONS` constants were converted to `defineDependencies()` / `defineSingletons()`, matching `HttpBootloader`, `RouterBootloader` and `FiltersBootloader`.

## Why

**1. Applications can now decorate `MailerInterface`.**

This is the main motivation. All bootloader bindings are registered before any `init()`/`boot()` method runs (`Initializer::init()` is fully materialised in `DefaultInvokerStrategy`). Because of that, the `bindSingleton()` call in `MailerBootloader::boot()` used to overwrite *any* `MailerInterface` binding declared by an application bootloader, no matter the bootloader order — decorating the mailer was effectively impossible without re-binding from a custom `boot()`.

With the binding declared in the singleton map, normal last-registration-wins semantics apply, and a decorator can depend on the concrete `MailQueue`:

```php
final class LoggingMailer implements MailerInterface
{
    public function __construct(
        private readonly MailQueue $next,
        private readonly LoggerInterface $logger,
    ) {}

    public function send(MessageInterface ...$message): void
    {
        $this->logger->info('Queueing {count} message(s)', ['count' => \count($message)]);
        $this->next->send(...$message);
    }
}

final class AppMailerBootloader extends Bootloader
{
    public function defineSingletons(): array
    {
        return [MailerInterface::class => LoggingMailer::class];
    }
}
```

**2. `MailQueue` is resolvable from the container — on the right queue connection.**

Previously `MailQueue::class` had no binding, so the container fell back to autowiring it. Its `QueueInterface` dependency was then resolved through `QueueInjector` without any context, i.e. on the **default** connection — the `queueConnection` config option was silently ignored. The only object honouring that option was the one built by the closure in `boot()`, and it was reachable exclusively as `MailerInterface`.

Now both aliases resolve to the same singleton, built on the configured connection.

## BC

`MailerInterface` still resolves to a `MailQueue` singleton, just registered in an earlier phase. A bootloader registered *before* `MailerBootloader` was overwritten previously (by `boot()`) and is overwritten now (by `defineSingletons()`) — same outcome.

Two things to be aware of, since `MailerBootloader` is intentionally non-final (#683):

- `boot()` lost its now-unused `BinderInterface $binder` argument. A subclass overriding `boot()` with the old signature becomes incompatible.
- A subclass redeclaring `protected const SINGLETONS` / `DEPENDENCIES` is now silently ignored, because the base class no longer reads those constants for this bootloader. Such subclasses should override `defineSingletons()` / `defineDependencies()` instead.

`initMailQueue()` is `protected` rather than `private`, so the queue construction stays overridable like the other factory methods.

## Tests

Full suite passes. The single failure on `Core\Internal\Proxy\ProxyClassRendererTest::testRenderParameter` (data set #20) reproduces on an unmodified `master` under PHP 8.4 and is unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/SendIt/src/Bootloader/MailerBootloader.php
+++ b/src/SendIt/src/Bootloader/MailerBootloader.php
@@ -37,6 +37,8 @@ class MailerBootloader extends Bootloader
     ];
     protected const SINGLETONS = [
         MailJob::class => MailJob::class,
+        MailQueue::class => [self::class, 'initMailQueue'],
+        MailerInterface::class => MailQueue::class,
         SymfonyMailer::class => [self::class, 'mailer'],
         TransportResolver::class => [self::class, 'initTransportResolver'],
         TransportResolverInterface::class => TransportResolver::class,
@@ -58,16 +60,8 @@ class MailerBootloader extends Bootloader
         ]);
     }
 
-    public function boot(BinderInterface $binder, ContainerInterface $container): void
+    public function boot(ContainerInterface $container): void
     {
-        $binder->bindSingleton(
-            MailerInterface::class,
-            static fn(MailerConfig $config, QueueConnectionProviderInterface $provider): MailQueue => new MailQueue(
-                $config,
-                $provider->getConnection($config->getQueueConnection()),
-            ),
-        );
-
         $registry = $container->get(QueueRegistry::class);
         \assert($registry instanceof QueueRegistry);
         $registry->setHandler(MailQueue::JOB_NAME, MailJob::class);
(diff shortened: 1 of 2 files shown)


roxblnfk:
Pushed 616e02f: moved the two new bindings back into the `SINGLETONS`/`DEPENDENCIES` constants instead of `defineSingletons()`/`defineDependencies()` overrides.

Reason: `MailerBootloader` is intentionally non-final (#683), and the base `Bootloader` resolves `static::SINGLETONS` — hardcoded `define*()` methods would silently ignore constant overrides in downstream subclasses. Both new entries are valid constant expressions, so the method form wasn't needed; the constant form also matches `BuilderBootloader`/`QueueBootloader`.

roxblnfk (approved):

